PR #10, opened 2026-04-24
Ship Load Object Respawning

TheRealMasterChief117:
## About the PR
During shipload, certain objects like ship generators, AME controllers, and grav gens will be respawned
Ships also now unshield upon load
AME controllers will refill aswell upon load

## Technical details
Minor namespace fixes to some files

## Media
Screenshot_20260424_143047

Ship has gravity and the shield generator no longer spams the sound

**Changelog**
:cl:
- fix: Fixed loaded shield generators spamming the shield dissipate sound
- fix: Maybe fix AME controllers breaking on ship load
- fix: Fixed loaded ships starting with no gravity


DDrakov:
Looks good, merging as this is needed badly

PR #25, opened 2026-05-02
Ship Save Changes: SRD Now Works on Loaded Ships, Performance Enhancements, Directional Message

TheRealMasterChief117:
## About the PR
- Loaded ships now give the loader a directional message (e.g. your ship is located 25 meters to the northwest!)
- Performance enhancements for saving, it no longer iterates through every entity with the SaveContrabandComp or whatever, everything is checked inside the invalid entity list instead
- SRD now generates repair data on ship load
- Entities inside of nested containers will correctly delete contraband now

## Media

Screenshot_20260502_151139
Screenshot_20260502_122947

Screenshot_20260502_154902

## Technical details
``HLSavedVesselPrototypeComponent`` has been deprecated. Monolith's ``VesselComponent``, does the exact same thing as this. The component wasn't removed code-wise so ships with this component don't become evil, but newly saved ships will not have this component.

New ``ShipyardShuttleLoadEvent``

**Changelog**
:cl:
- fix: Ships can no longer be saved if there are any mobs onboard.
- tweak: SRD now works on loaded ships. Any SRD data will be overwritten upon save, so make sure you repair your ship before you save!
- tweak: Loading a ship will now give you the relative direction of your ship, like purchasing.
- tweak: Medical stashes can now fit medipen and syringe cases


ColdDead:
Can you add the combat cases to? is legit sringe case just diffrent color

PR #44, opened 2026-05-09
Atmos Changes

TheRealMasterChief117:
## About the PR
Ports a few PRs from Frontier:
- https://github.com/new-frontiers-14/frontier-station-14/pull/3061
- https://github.com/new-frontiers-14/frontier-station-14/pull/3116

The gas deposit scanner change is simply for QoL.
For disabling atmos siphoning on exped planets, it is currently way too overpowered to keep in the game. There isn't any challenge or gameplay behind it, all you need to is set up a small/medium sized chamber with some scrubbers and you will make passive money (25k up to almost 125k on nitrous oxide/co2/water vapor planets)

Gas siphoning planets can be added back if we get extra gameplay mechanics behind it so it isn't just boring free money

**Changelog**
:cl: Frontier-Station
- remove: You can no longer siphon gas onto your ship from expedition planets.

### DIFF
--- a/Content.Server/Atmos/EntitySystems/AirFilterSystem.cs
+++ b/Content.Server/Atmos/EntitySystems/AirFilterSystem.cs
@@ -31,9 +31,12 @@ public sealed class AirFilterSystem : EntitySystem
         if (air.Pressure >= intake.Pressure)
             return;
 
-        var map = _atmosphere.AllowMapGasExtraction ? args.Map : null; // Frontier
+        // Frontier: check running gas extraction
+        if (!_atmosphere.AtmosInputCanRunOnMap(args.Map))
+            return;
+        // End Frontier
 
-        var environment = _atmosphere.GetContainingMixture(uid, args.Grid, map, true, true); // Frontier: args.Map<map
+        var environment = _atmosphere.GetContainingMixture(uid, args.Grid, args.Map, true, true);
         // nothing to intake from
         if (environment == null)
             return;
--- a/Content.Server/Atmos/EntitySystems/AirFilterSystem.cs
+++ b/Content.Server/Atmos/EntitySystems/AirFilterSystem.cs
@@ -31,7 +31,9 @@ public sealed class AirFilterSystem : EntitySystem
         if (air.Pressure >= intake.Pressure)
             return;
 
-        var environment = _atmosphere.GetContainingMixture(uid, args.Grid, args.Map, true, true);
+        var map = _atmosphere.AllowMapGasExtraction ? args.Map : null; // Frontier
+
+        var environment = _atmosphere.GetContainingMixture(uid, args.Grid, map, true, true); // Frontier: args.Map<map
         // nothing to intake from
         if (environment == null)
             return;
--- a/Content.Server/Atmos/EntitySystems/AtmosphereSystem.API.cs
+++ b/Content.Server/Atmos/EntitySystems/AtmosphereSystem.API.cs
@@ -167,12 +167,6 @@ public partial class AtmosphereSystem
             && Resolve(gridEnt, ref gridEnt.Comp1, false)
             && gridEnt.Comp1.Tiles.TryGetValue(gridTile, out var tile))
         {
-            // Frontier: hack to avoid infinite gas extraction from planets
-            if (tile.MapAtmosphere && map == null)
-            {
-                return GasMixture.SpaceGas;
-            }
-            // End Frontier
             if (excite)
             {
                 AddActiveTile(gridEnt.Comp1, tile);
--- a/Content.Server/Atmos/EntitySystems/AtmosphereSystem.API.cs
+++ b/Content.Server/Atmos/EntitySystems/AtmosphereSystem.API.cs
@@ -167,6 +167,12 @@ public partial class AtmosphereSystem
             && Resolve(gridEnt, ref gridEnt.Comp1, false)
             && gridEnt.Comp1.Tiles.TryGetValue(gridTile, out var tile))
         {
+            // Frontier: hack to avoid infinite gas extraction from planets
+            if (tile.MapAtmosphere && map == null)
+            {
+                return GasMixture.SpaceGas;
+            }
+            // End Frontier
             if (excite)
             {
                 AddActiveTile(gridEnt.Comp1, tile);
--- a/Content.Server/Atmos/EntitySystems/AtmosphereSystem.CVars.cs
+++ b/Content.Server/Atmos/EntitySystems/AtmosphereSystem.CVars.cs
@@ -1,3 +1,4 @@
+using Content.Shared._NF.CCVar; // Frontier
 using Content.Shared.CCVar;
 using Robust.Shared.Configuration;
 
@@ -26,6 +27,7 @@ namespace Content.Server.Atmos.EntitySystems
         public float AtmosTickRate { get; private set; }
         public float Speedup { get; private set; }
         public float HeatScale { get; private set; }
+        public bool AllowMapGasExtraction { get; private set; } // Frontier
 
         /// <summary>
         /// Time between each atmos sub-update.  If you are writing an atmos device, use AtmosDeviceUpdateEvent.dt
@@ -55,6 +57,7 @@ namespace Content.Server.Atmos.EntitySystems
             Subs.CVar(_cfg, CCVars.AtmosHeatScale, value => { HeatScale = value; InitializeGases(); }, true);
             Subs.CVar(_cfg, CCVars.ExcitedGroups, value => ExcitedGroups = value, true);
             Subs.CVar(_cfg, CCVars.ExcitedGroupsSpaceIsAllConsuming, value => ExcitedGroupsSpaceIsAllConsuming = value, true);
+            Subs.CVar(_cfg, NFCCVars.AllowMapGasExtraction, value => AllowMapGasExtraction = value, true); // Frontier
         }
     }
 }
--- a/Content.Server/Atmos/EntitySystems/AtmosphereSystem.Utils.cs
+++ b/Content.Server/Atmos/EntitySystems/AtmosphereSystem.Utils.cs
@@ -4,6 +4,7 @@ using Content.Server.Maps;
 using Content.Shared.Atmos;
 using Content.Shared.Atmos.Components;
 using Content.Shared.Maps;
+using Content.Shared.Shuttles.Components; // Frontier
 using Robust.Shared.Map;
 using Robust.Shared.Map.Components;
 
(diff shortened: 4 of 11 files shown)


Triad-Sector:
Okay, buff the deposits size then and test if it actually works idr if we fixed gas roid mining or not.

Erol509:
People prefer using liquid oxygen than what you can actually get from these expeds. Rest is quite useless

Triad-Sector:
Also we dont have plasma or other exotic gases on exped so Idk if atmos sucking would be good

TheRealMasterChief117:
Nitrous, water vapor, and carbon dioxide sell for a bit
It's also just free money either way

Triad-Sector:
they dont sell for that much imo Nitrium sells for way more

rebaserHEAD:
This is more so a change due to the main bottleneck in frezon production being oxygen and nitrogen for trit production and again for frezon respectively. Both are found planet side in abundance. This negates the need for a mining crew to seek out these materials and instead can go and run an expedition to hoover up massive quantities of the materials.

That's largely the reason for this PR, I believe.

TheRealMasterChief117:
and for mass-selling a large amount of infinite gas from exped planets

> This is more so a change due to the main bottleneck in frezon production being oxygen and nitrogen for trit production and again for frezon respectively. Both are found planet side in abundance. This negates the need for a mining crew to seek out these materials and instead can go and run an expedition to hoover up massive quantities of the materials.
> 
> That's largely the reason for this PR, I believe.

And for selling a large amount of siphoned gas from exped planets
I have made around 100-200k by doing nothing except siphon nitrous, carbon dioxide, and ammonia per 1-2 expeds
Very large amount of money for simply waiting 15 minutes, and with nothing to actually sacrifice besides from making a setup and forget chamber

TschudyNF:
Personally voting "hell no" on this since i use expedatmo to reoressurize my ship in a way that doesnt necessitate being parked at CC for an hour and a half. 

If the issue is that its too lucrative for its difficulty in current state, then cut gas sale prices by 90% or so until it can be better balanced.

DDrakov:
The issue is mostly just how easy you can get some gas from exped. If we're talking about re-pressurize atmos for ships MD has air canister as well as other connector too like oxy/nitrogen if people need it.

TschudyNF:
That can take several minutes up into hours depending on the size of ship.  If the issue is that the base gasses are too lucrative, then slash the price of the base gasses.

TheRealMasterChief117:
Air grenades were buffed and expedite the process of refilling a ship immensely

TheRealMasterChief117:
> That can take several minutes up into hours depending on the size of ship.  If the issue is that the base gasses are too lucrative, then slash the price of the base gasses.

It's not even "base gasses"

It's ammonia, nitrous, and carbon dioxide all being lucrative gasses, especially when the atmos is fully comprised of it

TschudyNF:
1 cred per mol.  still passive money but nowhere near as much as if they actually process it into other gases.

TheRealMasterChief117:
I have been able to make a consistent 150k per expedition by simply picking dioxide, ammonia, or nitrous oxide expeds

It is a crazy amount of money for 15 minutes of doing nothing

Th3ShadowR3ealm:
I agree with others here. Exped mining is too high, but the solution should be changing the base sell prices of exped spawned gasses, not making atmos changes.

Th3ShadowR3ealm:
> The issue is mostly just how easy you can get some gas from exped. If we're talking about re-pressurize atmos for ships MD has air canister as well as other connector too like oxy/nitrogen if people need it.



> Personally voting "hell no" on this since i use expedatmo to reoressurize my ship in a way that doesnt necessitate being parked at CC for an hour and a half.
> 
> If the issue is that its too lucrative for its difficulty in current state, then cut gas sale prices by 90% or so until it can be better balanced.

If we stop at md or cc to repressurize after every issue there won't EVER be docking space. Stc will be a nightmare.

TheRealMasterChief117:
> I agree with others here. Exped mining is too high, but the solution should be changing the base sell prices of exped spawned gasses, not making atmos changes. 

This would require some finicky code, might be possible unsure

SquishyEmmaaaa:
I think having the gameplay loop for gases be from asteroids is a good change.

Otherwise, theres no downside to a team who is running expeds just turning an exped hull into an atmos harvester and double dipping on income, with the harvesting of atmos just being free money in the background.

Dimasw99:
I am working on ideas of how to tweak and adjust gas mining to be more lucrative though it ia a lot of apread sheet work right now.

About the money, 600k per hour is funnily enough one of the more tame methods and is quite close to just basic mining with the best upgrades.

There are currently more significant sources of stupid amounts of wealth like drone hunting yielding 3mil an hour or more and the crystalizer making 8000 plastitanium (that number isn't a figure of speech and a real number)

Edit:
A good way to nerf exped gasses is to use nitrogen. Simply adjust the ratio between high value gas to low value and you got a play strategy that ia viable but not game breaking. Which I think is the most important part since removing play methods is boring

TheRealMasterChief117:
Agreed, gas mining from expeditions should still be possible but definitely less lucrative

Eillyrea:
Gas mining from expeditions shouldn't be a thing.

It shouldn't exist that an expedition ship is inherently better at getting a lot of oxygen than a dedicated atmospherics ship.  When oxygen is in bulk needed for a lot of atmospherics work.

The windreign is a dedicated atmos ship, that have extra ships off of it for gas mining and cannot go on expeditions so will always be behind an exped ship that can land, suck up all the air and unland as much as they want.

Draining atmospheres from planets is unintended gameplay and should be removed.

TschudyNF:
The idea that siphoning atmo gasses from a planet is "unintended gameplay" simply doesn't hold water.  Its no secret that mono's lack of refinement on the gas mining mechanics leaves it in a less desirable state than it should be.  Im not arguing that.  But this means that proper gas mining; the kind where you clear an asteroid, set up drills, hook up pipes, and drain that thing dry needs to be fixed and improved.   

A proper gas miner absolutely should be able to outperform any other class of ship when it comes to that task, but that can't be used as an excuse to disable it for everyone else while leaving the gameplay loop in a crappy state because its not high enough on the priority list.  If it doesn't need to be fixed *now*, then it doesn't need to be disabled *now*.

This is another example of something that's creating a troubling pattern.  This mentality where  the current state if something doesn't fit the grand design and warrants *immediate* removal, but at the same time is so insignificant that taking any steps toward an ideal state its tossed into the priority void of "if someone gets around to it"...just feels so incredibly toxic.

TheRealMasterChief117:
Draining gas from planets is not unintended gameplay

It is entirely intentional
The reason why this was brought up was because you can make 200k+ for free on ammonia, nitrous oxide, and carbon dioxide expedition fairly easy.

What we're going to do now, likely, is not disable it, but reduce the rates so it's still a decent passive way to get gas, but asteroid gas mining will always out class it

Eillyrea:
> Draining gas from planets is not unintended gameplay 
It is entirely intentional

No, atmos scrubbers and expedition ships were not made with the idea in mind that both of them could be used together to drain planets of their atmos by frontier forks.  Surprisingly, that was never how they were designed.

To quote the PR you literally linked: "Until we have a proper solution where planetary atmospheric collection is challenging, and an intentional design that makes it more engaging, this removal addresses the issue of infinite resource extraction."

There's a reason it's been disabled on both Coyote and Frontier, and it's because expedition ships shouldn't be functioning better than atmospheric ships doing their intended gameplay loop.  It's not just amount ammonia/carbon and nitrous oxide but O2 as well.

TheRealMasterChief117:
increase the amount of nitrogen per exped, and you get reduced O2 rates as well
The problem at hand here is expedition ships are faster at gaining gas and credits then dedicated gas miners, which can be alleviated. I don't believe we should be removing ways to obtain gases. It may be unintentional on Coyote and Frontier, but not here. There's a reason why this PR is under discussion, too, and I have given it some thought, and as I said earlier it sparked because I was able to make hundreds of thousands of credits per exped. 

we could:
- Increase nitrogen per exped, making it so high value gases are not as prominent
- Force scrubber rates to lower while on a planet map (and passive vents)
- Make high-O2 planets non-existent
- maybe even make the amount of gas extractable finite

TschudyNF:
Would be completely fine with all 4 of those put together.  Im not a atmo main, im not going to claim that unlimited access to o2 and n2 can't be mcguyvered into scrooge macduckian wealth, nor that a dedicated gas miner should not outmatch a few planetside scrubbers by several magnitudes.

I can't speak for others, but my system is designed to refill our normal breathing atmosphere while running expeditions, while maybe filtering a third into cans.

Dimasw99:
Removing a gameplay method or loop should only be a last resort when all other options fail and it should be considered how detrimental it is for a server. In my eyes when I see that exped gas extraction is better than normal asteroid mining then you also need to ask the question if maybe gas mining from asteroids isn't too weak. Extracting gas from a planet is not free after all, it takes player time which is also a very valuable resource to consider. If a player is able to multitask and benefit from it then that is great and smart play.

Likewise it is irrelevant to say if something was intentional or not. What matter is simply, is it fun? Yes? Then it should stay. Same goes for looking at other forks. Only because they do it doesn't mean we have to follow.

I think the real big problem we have as a server is that there isn't really a set in stone guideline for how much a player should earn per hour, or atleast that I know of. A design document would help immensily which should be held in an open discussion. 

Anyways I don't consider this gameplay loop and method to be busted in any degree compared to other things floating about.

Eillyrea:
> increase the amount of nitrogen per exped, and you get reduced O2 rates as well

This negatively impacts people who just want to walk around on a planet without needing to use their internals so that a few atmos powergamers can powergame

> It may be unintentional on Coyote and Frontier, but not here. There's a reason why this PR is under discussion

It was never 'intentional' here, it was pulled in as part of our upstream. Wherein it becomes up for discussion and is default unintentional as it is in our upstreams

> Increase nitrogen per exped, making it so high value gases are not as prominent
    Force scrubber rates to lower while on a planet map (and passive vents)
    Make high-O2 planets non-existent
    maybe even make the amount of gas extractable finite

None of these proposed solutions fix the fact that expedition ships have something they can do to passively acquire gas whilst an atmosian is working on their systems that a atmos ship cannot do.

> maybe gas mining from asteroids isn't too weak.

It is, but ships having the capacity to put down 50x80 tiles worth of "pull up the atmosphere" setup.  The most egregious of which I've seen is 72 scrubbers on a ship, is too much.  The intended gameplay loop shouldn't be "Click a button for unlimited gas in your own private instance away from the rest of the server"

> Likewise it is irrelevant to say if something was intentional or not. What matter is simply, is it fun? Yes? Then it should stay.

I don't know what part of clicking the funny expedition button and getting infinite resources is fun...  I prefer gameplay loops where I'm actively doing things.  Unless you're enjoying expeditions and just see getting infinite gas as a nice side bonus, which isn't what expeds are designed for.

> Anyways I don't consider this gameplay loop and method to be busted in any degree compared to other things floating about.

Other methods of play are more busted, yes.  Drone farming being one of them.  But just because X is worse, does not mean we ignore the problem of Y.  We have a solution to Y so it should be implemented.  People converting expedition ships into atmos ships to drain the atmosphere of a planet is clearly not intended otherwise:
A) An atmospherics ship would be able to do it
B) Other forks wouldn't have disabled it
C) People wouldn't be having to go out of their way to create ships with specific components that can do it (the capacity to expedition)
D) It would say anywhere in the guidebook that this is something that can be done

It is plain and simply, unintended combination of mechanics.  Whilst I'm all for offshore gas mining to be buffed and more viable of a technique for gameplay, that doesn't mean we shouldn't remove things that are soley there to benefit very few atmos powergamers who go build specialised expedition ships for atmosphere sucking.

TschudyNF:
It doesn't mean that we should remove it either.  The normal atmo ship loop doesn't stop existing or actively get worse because a single digit of try-hards makes their number go up faster.

It also wont get any better if the balancing focus remains on robbing Peter to make Paul feel less poor by comparison.

PR #77, opened 2026-05-14
Allows Reptilians to Have Hair

mnva0:
## About the PR
This PR allows Reptilians to have hair ~~and facial hair.~~ The way I did this *should be* completely identical to the way DeltaV did it

Update: Facial hair has been removed

## Why / Balance
This was briefly brought up in the contributing channel and was met with resounding apathy by maintainers \j

Some of the people porting their Reptilian characters here are likely from servers *with* Reptilian hair (the biggest I can immediately think of is DeltaV, though I am unsure of how many players we have from there), so this'll allow them to keep their characters closer to how they were. Hair is a nice way to characterize and accessorize, and people who simply Do Not want Reptilian hair for their characters can just not add hair. I dislike the idea of people needing to downgrade their characters to play here or just... not being able to customize their characters the way they want to

There *are* two problems with this, but I believe they're minor enough to not really be *that* much of a concern
- ~~There are Reptilian markings that are supposed to look like some hairs. This isn't a big incompatibility and the selection is very limited so whatever. The only thing that *might* be an issue is an eventual removal of those markings? I'm not sure how cleanly the game can handle something like that, but I *imagine* it'll just disappear or something~~ I can't... actually find them in the hair menu? Okay, maybe those are exclusively Reptilian. Sure, why not. The hair-like markings in question are: Burqa, Hijab (both variants), Niqab, Tichel. I think I saw these as hairs on another server.
- Cryo pod bodies might have some *ghastly* cuts, but this is also ehhhh. ~~If you hate their cut, just include fixing a hair emergency as part of the medical services~~

I'll maybe look into a solution for these later because they aren't exactly gamebreaking. For removing hair from the emergency pod bodies, it's likely related to RandomAppearance which has a "Hair" field

## Media
image

image

## Requirements

## How to test
1. Get a reptilian character
2. Add hair

**Changelog**
:cl: Minerva
- tweak: Reptilians can now have hair, though not facial hair.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Species/reptilian.yml
+++ b/Resources/Prototypes/Entities/Mobs/Species/reptilian.yml
@@ -8,6 +8,7 @@
   - type: HumanoidAppearance
     species: Reptilian
     hideLayersOnEquip:
+    - Hair # Triad - Reptilian hair.
     - Snout
 #    - HeadTop # WWDP - Because visible horns its cool
     - HeadSide
@@ -165,6 +166,7 @@
   - type: HumanoidAppearance
     species: Reptilian
     hideLayersOnEquip: # WWDP
+    - Hair # Triad - Reptilian hair.
     - Snout
 #    - HeadTop # WWDP - Because visible horns its cool
     - HeadSide
--- a/Resources/Prototypes/Species/reptilian.yml
+++ b/Resources/Prototypes/Species/reptilian.yml
@@ -16,6 +16,8 @@
   id: MobReptilianSprites
   sprites:
     Head: MobReptilianHead
+    Hair: MobHumanoidAnyMarking # Triad - Reptilian hair.
+    FacialHair: MobHumanoidAnyMarking # Triad - Reptilian hair.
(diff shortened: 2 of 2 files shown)


DDrakov (approved):
I'm alright with this, need another maintainer input

TheRealMasterChief117 (changes requested):
I'll need to think on this, it does fuck with the overall style of reptilians.
As for randomly generated reptilians, I can code something to prevent that and add it to this PR after I do some thinking.

TheRealMasterChief117 (changes requested):
Make reptilians unable to have facial hair and we're good

mnva0:
Facial hair has been removed from Reptilians as of the last commit

image


This keeps the broken behavior described in https://github.com/Triad-Sector/Triad_Sector/issues/165 ("Various species have hair/facial hair slots which do not work") because the alternative might be whitelisting every hair for Reptilian, which would take... a very long time and might be far more trouble than it's worth and every hair option ported might have to be made compatible

TheRealMasterChief117 (approved):

PR #78, opened 2026-05-14
Alternate Backpack Solution

TheRealMasterChief117:
## About the PR
Backpack is now smaller again, reverted back to 4x4
As a result you can only put 1 backpack per stash
Allows people to carry important items alongside a backpack without allowing the hoarding of tactical backpacks

## Technical details
Ports RMC limited storage system

## Media
image

**Changelog**
:cl:
- tweak: Tactical backpacks are now 4x4 again.
- tweak: You can only fit 1 tactical backpack into a military stash,


DDrakov (approved):
it looks good, should be good to go

PR #85, opened 2026-05-15
Lights on Prebuilt Light Fixtures Aren't Deleted on Save

### LINKED ISSUE #53: Lights don't seem to be persistent
## Description
Light fixtures apparently do not save, meaning light swaps for existing fixtures (important for building!) are temporary. Additionally, removed lights replace themselves meaning you'd likely have to remove a fixture entirely if you want it to remain empty

Aside: I'm not sure how much I like the lights automatically respawning as a mechanic for convenience. I thought about porting over shielded lights from... maybe Frontier or Wizden—not sure where—as a theoretical way to reduce maintenance. Also provides a bit more progression for ship upgrades, too

If I had to guess, `ShipyardGridSaveSystem.cs` might have a solution. It does have a dedicated part for `PoweredLight`

**Reproduction**
1. Get a shuttle
2. Remove some light tubes/bulbs. Maybe swap some. Maybe add some
3. Save and load the shuttle.
4. Your swaps won't save.  New lights should be kept as-is

**Screenshots**

**Additional context**
My saving and loaded was done on the same round on the Gondola

I tried fixing this before in the PR commented below

TheRealMasterChief117:
## About the PR
Title
Resolves #53 

Meaning lights aren't replaced on shipload from prebuilt fixtures, meaning custom lights will stay

## Breaking changes

**Changelog**
:cl:
- fix: Fixed lights inside of prebuilt fixtures being replaced on ship load, meaning your custom lights will save correctly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -313,3 +313,5 @@ Resources/MapImages
 # Link Files
 *.lnk
 
+# lscache
+*.lscache
--- a/Content.Server/Light/Components/PoweredLightComponent.cs
+++ b/Content.Server/Light/Components/PoweredLightComponent.cs
@@ -6,13 +6,14 @@ using Robust.Shared.Audio;
 using Robust.Shared.Containers;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Prototype;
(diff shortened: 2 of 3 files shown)


DDrakov (approved):
All is good, need to speedmerge as well due to .gitignore

PR #87, opened 2026-05-15
Triad Lore in the Guidebook

TheRealMasterChief117:
## About the PR
Lore
Adds wizden fixes to guidebook hyperlink

TODO:
- [x] Fix hyperlinks in the guidebook
- [x] Add the lore primer
- [x] Add a page for each major faction
- [x] Add the full timeline
- [x] Change/remove references of disabled companies
- [x] Change references to Colossus to Triad
- [x] Final touchups to the lore

Lore is by Sigmund, with the port to the guidebook by me! Thank you mr. Loretainer

Disabled these companies due to non-compliance with the lore:
- Dark Matter Enterprises
- CDM
- Nosske-Eien
- Blackhawk PMC

These companies can be added back if they are given new lore and perhaps some more love
CDM is removed due to similarities with the CDF, which is what the Triad Defense Force used to be called before it was renamed.

## Media
Also adds a date in the PDA
image

image
image

**Changelog**
changelog part in the file to account for the two names

PR #103, opened 2026-05-20
Cherrypick: Add a Countdown to the PDA Showing Time Remaining Until Roundend (#4147)

mnva0:
## About the PR

This PR cherrypicks https://github.com/new-frontiers-14/frontier-station-14/pull/4147, which adds a timer to the PDA that says how long until the round ends

From the original PR:
> Adds a field to the PDA underneath the current shift time showing how long until the round will end.
Mostly accurate (it doesn't take into account the extended shuttle call time if the current alert level is delta or epsilon. But in those cases, it's unlikely that it would be those alert levels for an extended period of time, and the error is only 10 extra minutes of shift time)

I don't know how to test this in an environment more like the server, but given Frontier has this on live it's *probably* fine. It's cvar-based anyway, so I expect it to work fine
The clock reads 10 minutes left at the same time (like off by a couple seconds but whatever) the 10 minute warning announcement is given

## Why / Balance
It's REALLY useful knowing when exactly the round will end, and it's especially useful for people who aren't sure

## Media
image

image

## Requirements

## How to test
From the original PR:

> Use the command > cvar shuttle.auto_call_time to set the time until round end (if you had your PDA open before this, reopen the menu)
Check that the PDA matches that + 10 minutes
Call the shuttle and check that the PDA matches the shuttle countdown
Let the countdown hit 0 and see that it disappears when the round end summary window appears.

**Changelog**
:cl: Frontier
- add: Added a countdown to when the shift will end to the PDA. Hey! Stop watching the clock and get back to work!

### DIFF
--- a/Content.Client/PDA/PdaMenu.xaml
+++ b/Content.Client/PDA/PdaMenu.xaml
@@ -44,6 +44,11 @@
             <ContainerButton Name="StationTimeButton">
                 <RichTextLabel Name="StationTimeLabel" Access="Public"/>
             </ContainerButton>
+            <!-- Frontier -->
+            <ContainerButton Name="RemainingTimeButton">
+                <RichTextLabel Name="RemainingTimeLabel" Access="Public"/>
+            </ContainerButton>
+            <!-- End Frontier -->
             <ContainerButton Name="StationAlertLevelInstructionsButton">
                 <RichTextLabel Name="StationAlertLevelInstructions" Access="Public"/>
             </ContainerButton>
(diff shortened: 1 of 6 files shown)


TheRealMasterChief117:
Conflicts with https://github.com/Triad-Sector/Triad_Sector/pull/87

DDrakov:
Curious if the time for shuttle to arrive is the same as the round end timer

mnva0:
> Curious if the time for shuttle to arrive is the same as the round end timer

From my memory (I have not tested specifically to remember this), it counts down the exact time until the roundend message pops up. Which should be the autocall time + 10 minutes
If the autocall time is set to 1 minute, at roundstart it'll read like 11 minutes until round end

Or at least I think so. It should behave the exact same way it does on Frontier

DDrakov:
> > Curious if the time for shuttle to arrive is the same as the round end timer
> 
> From my memory (I have not tested specifically to remember this), it counts down the exact time until the roundend message pops up. Which should be the autocall time + 10 minutes If the autocall time is set to 1 minute, at roundstart it'll read like 11 minutes until round end
> 
> Or at least I think so. It should behave the exact same way it does on Frontier

Huh, I'll test it on my devenv local later on

TheRealMasterChief117 (approved):

mnva0:
Selfmerge because Drakov said I could

PR #113, opened 2026-05-23
Disablers Deal No Damage

TheRealMasterChief117:
## About the PR
Wizden changed this years ago. Disablers being a non-lethal shouldn't be the MothKiller 3000.

## Technical details
Removes ``BulletDisablerStaminaOnly`` as a prototype. Frontier added this. Replaces all references of it with BulletDisabler.

**Changelog**
:cl:
- tweak: Disablers now deal 0 heat damage. As a result of this, pacifists can now use disablers.
- tweak: Practice weapons can now be used by pacifists.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -276,6 +276,7 @@
     fireCost: 30 #Mono: 62.5 >> 30
   - type: StaticPrice
     price: 300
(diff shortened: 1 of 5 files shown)


mnva0 (changes requested):
A lot of these line changes probably need some Triad comments? It's only commented sometimes

TheRealMasterChief117:
> A lot of these line changes probably need some Triad comments? It's only commented sometimes

The changes that aren't commented are those that are equal to Wizden

mnva0 (approved):
I'm not sure how to dismiss my requested change review so I'm trying this

PR #162, opened 2026-05-30
Ship Saving Sanitation

DDrakov:
## About the PR

Sanitize some of the needed change

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fix ironman loading ship bug, now's unable to load any ship.
- fix: Chemmaster bricked saving and loading.
- tweak: Fragment drone gun is unable to save due to it's instability and complicated mechanism.
- tweak: Loading a ship can no longer put you into debt, it will fail if you do not have enough funds.


IrkallaEpsilon:
Would it be possible to have a warning added to the contravends purchase of Emag + Auth disruptor. Theres already some folks that spend 40/50 TC on either of them which is now essentially wasted, it is basically avoiding future waste purchases.

Also would it be possible to remove them from the surplus crate loot pools as they are pretty much just another, slightly better, dummy drop with no ship saving for either?

SomeoneElsHere:
I was wondering, why is the free tax when you save and load quickly removed? Can you instead just remove it for getting a different ship?

DDrakov:
> Would it be possible to have a warning added to the contravends purchase of Emag + Auth disruptor. Theres already some folks that spend 40/50 TC on either of them which is now essentially wasted, it is basically avoiding future waste purchases.
> 
> Also would it be possible to remove them from the surplus crate loot pools as they are pretty much just another, slightly better, dummy drop with no ship saving for either?

I may update the FTL yeah, tho I won't remove them from loot pools since this can be exchanged in the future faction implementation for their own currencies.

DDrakov:
> I was wondering, why is the free tax when you save and load quickly removed? Can you instead just remove it for getting a different ship?

Perhaps, if I found a way I'll try to implement them, like saving and loading same ship within X amount of time.

Lazlo23:
> Perhaps, if I found a way I'll try to implement them, like saving and loading same ship within X amount of time.



i wonder if you could do it by checking if the second shuttle loaded is equal or less than the one you saved recently, then if it is equal or lesser value, you can load it free?

DDrakov:
So about the ship contraband on some items
1. Technically it's syndicate item and it's access breaker + it have a capacity to hack late and such so likely shipyard engineer going to confescate it
2. I'll think about trying to apply the no tax for now but shouldn't be issue as in the future I'm planning to add more economy source and rebalance.

Content.Server/_NF/Shipyard/Systems/ShipyardSystem.Consoles.cs
```diff
@@ -816,6 +807,30 @@ public void OnLoadMessage(EntityUid uid, ShipyardConsoleComponent component, Shi
...
+            companyName = voucherCompany.CompanyName;
+        }
+
+        // Apply company to ship if we found one
+        if (!string.IsNullOrEmpty(companyName))
+        {
+            var shipCompany = EnsureComp<CompanyComponent>(shuttleUid);
+            shipCompany.CompanyName = companyName;
+            Dirty(shuttleUid, shipCompany);
+        }
```
TheRealMasterChief117: Should be a generic method to set the company name so that copy-paste isn't present with purchasing as well

Content.Server/_NF/Shipyard/Systems/ShipyardSystem.cs
```diff
@@ -449,14 +449,14 @@ private void AddShipAccessToEntities(EntityUid gridUid)
...
             {
-                EnsureComp<ShipAccessReaderComponent>(entity);
+                EntityManager.EnsureComponent<ShipAccessReaderComponent>(entity);
             }
             // Add ship access to entity storage (lockers, crates, etc.)
-            else if (HasComp<EntityStorageComponent>(entity))
+            else if (EntityManager.HasComponent<EntityStorageComponent>(entity))
             {
-                EnsureComp<ShipAccessReaderComponent>(entity);
+                EntityManager.EnsureComponent<ShipAccessReaderComponent>(entity);
```
TheRealMasterChief117: How come all of this is changed?

Content.Server/Chemistry/EntitySystems/ChemMasterSystem.cs
```diff
@@ -65,6 +68,20 @@ public override void Initialize()
...
+            // Log the buffer contents for debugging
+            if (_solutionContainerSystem.TryGetSolution(ent.Owner, SharedChemMaster.BufferSolutionName, out _, out var bufferSolution))
+            {
+                Logger.Info($"ChemMaster {ent.Owner} loaded with buffer: {bufferSolution.Volume}u, {bufferSolution.Contents.Count} reagent types");
+                foreach (var reagent in bufferSolution.Contents)
+                {
+                    Logger.Info($"  - {reagent.Reagent.Prototype}: {reagent.Quantity}u");
+                }
+            }
+        }
```
TheRealMasterChief117: This'll spam logs everytime a chem master gets loaded, do we want this?

Content.Server/_NF/Shipyard/Systems/ShipyardSystem.Consoles.cs
```diff
@@ -920,10 +944,14 @@ public void OnLoadMessage(EntityUid uid, ShipyardConsoleComponent component, Shi
         }
         if (shuttleStation != null)
             _records.Synchronize(shuttleStation.Value);
+            _records.Synchronize(station);
```
TheRealMasterChief117:
```suggestion
if (shuttleStation != null)
    _records.Synchronize(shuttleStation.Value);

_records.Synchronize(station);
```
Not inside of this if statement, so should be indented correctly

Resources/Prototypes/_NF/Entities/Objects/Devices/pinpointer.yml
```diff
@@ -19,7 +19,7 @@
     component: SyndicateFOB
     targetName: unknown
   - type: SavingContraband # Triad
-    examineText: ship-saving-contraband-1-text
+    examineText: ship-saving-contraband-2-text
```
TheRealMasterChief117:
```suggestion
    examineText: ship-saving-contraband-1-text
```
Keep this as one if the intent is for an item to be seized on save but not illegal to own
Two implies that its illegal and will be confisicated; we don't want TDF players thinking that this item should be seized from players.

Content.Server/_NF/Shipyard/Systems/ShipyardSystem.Consoles.cs
DDrakov: Yeah, prob gonna delete most of the things that I update on ShipyardSystem.Consoles.cs since there's already a better PR

Resources/Prototypes/_NF/Entities/Objects/Devices/pinpointer.yml
DDrakov: Yes it's intended, having it is fine but will be removed on load or saved

Content.Server/Chemistry/EntitySystems/ChemMasterSystem.cs
DDrakov: I'll check it back, but nah I don't want to spam

Resources/Prototypes/_NF/Entities/Objects/Devices/pinpointer.yml
TheRealMasterChief117: So uhh revert it, then?

DDrakov: Yeah I'll revert it when I came back to it

Content.Server/_Triad/Shipyard/ShipyardSystem.Triad.Consoles.cs
```diff
@@ -342,6 +342,14 @@ public void OnLoadMessage(EntityUid uid, ShipyardConsoleComponent component, Shi
             return;
         }
 
+        if (!_bank.TryBankWithdraw(player, appraisalCost))
+        {
+            Del(shuttleUid);
+            ConsolePopup(player, Loc.GetString("cargo-console-insufficient-funds", ("cost", appraisalCost)));
+            PlayDenySound(player, uid, component);
+            return;
+        }
```
rebaserHEAD, line 351: This guard is added in front of the existing withdrawal rather than replacing it, so the player pays twice: `TryBankWithdraw` here deducts and dirties the balance, then `TryBankWithdrawAllowDebt` a few lines below deducts the same `appraisalCost` again. (The `newBalance` calc below also reads `bankAccount.Balance` after this first deduction, so the debt popup is off too.)

Since we're dropping debt-loading, this guard plus a single success popup should replace the whole block from `var currentBalance` down through the debt/charged popup (that block sits just below this hunk, not shown here). It also swaps the borrowed cargo string for `shipyard-console-insufficient-funds`, which already exists and takes `cost` + `balance`:

```csharp
// Triad: no-debt withdrawal. Blocks ironman + insolvent players and deletes the
// orphaned grid on failure (the loaded shuttle was spawned above before charging).
if (!_bank.TryBankWithdraw(player, appraisalCost))
{
    Del(shuttleUid);
    ConsolePopup(player, Loc.GetString("shipyard-console-insufficient-funds",
        ("cost", appraisalCost), ("balance", bankAccount.Balance)));
    PlayDenySound(player, uid, component);
    return;
}

ConsolePopup(player, Loc.GetString("shipyard-console-load-success-charged",
    ("ship", name), ("cost", appraisalCost)));
```

That leaves `TryBankWithdrawAllowDebt` with no callers (dead code), worth commenting out with a reason in a follow-up.

TheRealMasterChief117 (approved):

PR #177, opened 2026-06-02
Test Improvements - Monolith Port

TheRealMasterChief117:
## About the PR
Ports https://github.com/Monolith-Station/Monolith/pull/3910 from monolith

This should fix Build & Run Shipyard tests failing randomly due to the dotnet version mismatch
There should be less warnings 

**Changelog**
NO player face


DDrakov (approved):
All looks good, checked the actions and the code function

PR #203, opened 2026-06-10
Crashes and Error-Flood Fix Roll Up

rebaserHEAD:
## About the PR
A batch of crash and error-flood fixes found by triaging the live server logs in Grafana/Loki and ranking error volume by subsystem. Most share one shape underneath: a deleted entity's id lingering in a field or collection and then used without an existence guard, plus a couple of empty-collection and missing-component cases. No gameplay or balance changes; each commit just restores intended behavior and removes a thrown exception or a per-tick log flood.

## Why / Balance
No balance impact, these are all bug fixes. Two dominated the picture: an ore silo handing a deleted silo's id to the PVS override system every tick (roughly five million logged errors a day) and a projectile anomaly indexing an empty target list on supercritical (roughly seventy thousand a day, which also left the supercritical effect broken). The rest are lower volume but real: a lathe throwing on recipe queue, auto-fire ballistic weapons throwing on a stale ammo reference, a DoAfter update that could abort the entity-system tick, and a few graceful-but-noisy Resolve errors (examine, clothing, vending state).

## Media
Us when logs.

image

## Requirements

## How to test
Build and boot a server; it should run a round without the error floods these target. Specific checks:
- **Lathe:** queue a recipe that needs a material the lathe currently has none of. It should queue without throwing (previously `KeyNotFoundException`).
- **Projectile anomaly:** spawn one away from any mobs/objects and push it to supercritical. It should not throw `ArgumentOutOfRangeException` every tick, and it should resolve normally.
- **Ballistic auto-fire:** auto-fire a ballistic weapon whose magazine had an ammo entity deleted out from under it. No `Entity 0 is not valid`.
- **Ore silo / vending:** over a normal round, the `non-existent entity 0 to PVS override` and `VendingMachinePurchase` PVS-serialize errors should be absent from the server log.

## Breaking changes
None to public APIs or prototypes. The only state-shape change: `VendingMachinePurchaseComponent` is no longer a `NetworkedComponent` (its two fields were `AutoNetworkedField`). It is read only server-side (cargo pallet pricing) and nothing on the client consumed it, so this is backward-safe; existing saved ships carrying the component still deserialize fine.

**Changelog**
:cl:
- fix: Fixed an error when queueing a lathe recipe for a material the lathe does not currently have stocked.
- fix: Projectile anomalies no longer fail to fire when they go supercritical with nothing in range.
- fix: Fixed an error that could occur while auto-firing ballistic weapons.
- fix: Improved server performance by eliminating several high-frequency internal errors (most notably from ore silos).


DDrakov (approved):
Looks good

PR #208, opened 2026-06-11
Adjust Guardian Damage From 0 to 2

Dimasw99:
## About the PR
Guardians used the practice laser to do damage which was set to 0 in #113 
Added a unique laser type for the Guardian doing 2 damage of the previous 1, buffing it.

## Why / Balance
A DPS of 8 meant that the guardian needed 12.5 seconds to kill an unarmored human if every shot hits. A hunter drone has a DPS of 17.6, killing a person in 5.7 seconds. Making it 16 DPS means that a player has 6.25 seconds of face tanking before dying unarmored.

Assuming a player is wearing a merc hardsuit which you can get from start it is now 9 seconds of standing still and surviving.
If a player is using a laser rifle of a DPS of 32 then the boss would die in 14 seconds.

These numbers of course are just cases where the player and boss both stand still and hit each other without strategy and moving. It should be a proper boss to fear now.

**Changelog**

:cl:
- tweak: Guardian Unit DPS buffed from 8 to 16
- fix: Guardian Unit doing 0 damage

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/hitscan.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/hitscan.yml
@@ -95,6 +95,16 @@
       types:
         Heat: 0 # Triad 1->0
 
+- type: entity  # Triad - With firerate 8, kill an unarmored person in 6.25 seconds
+  parent: BasicHitscan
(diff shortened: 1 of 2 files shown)


DDrakov (approved):
it's a good change, so guardian is not a shit anyway

PR #214, opened 2026-06-12
Make Bomber Jackets Insulative

TheRealMasterChief117:
## About the PR
new drip for our reptilian friends
The bomber jacket is a winter coat IRL, so it not being insulative here doesn't maky any sense

**Changelog**
:cl:
- tweak: Bomber jackets are now insulative, like winter coats


DDrakov (approved):
Really shrimple

PR #222, opened 2026-06-14
Ship Load Fee Is Now Visible in the Shipyard Console UI

TheRealMasterChief117:
## About the PR
Adds a ship load fee label to the shipyard console

The number is rounded to the nearest 2 significant digits because for some reason the appraisal prices keep changing across saves.. unsure why, but this works fine for now. It's only off by a couple hundred of credits

The number is not authoritative and is simply for UI so players can see an estimate
The server will gauge the actual price and WILL put you into debt

Also fixes the bug of the load button being correctly disabled when a ship isn't selected

## Why / Balance
So people can see how much the ship fee is worth before loading it!

## Media

Tax rate label is dynamic based on the CVAR
image

image

## Technical details
Makes ``TriadCCVars.LoadShipPrice`` ``SERVER` and ``REPLICATED`` so the client can see the value but not modify it
Minor XML changes

Adds the method ``GetKeyValueFromPath`` to ``ShipFileManagement`` system
This will allow you to get a certain mapping node from a ship YML file on the client  easily through a given string key

**Changelog**
:cl:
- tweak: An estimated appraisal of your ship will be displayed when you select a saved ship in the shipyard console. This value will not be visible if your ship was saved pre-tamper update. Resaving your ship will fix this, however.

### DIFF
--- a/Content.Client/_NF/Shipyard/BUI/ShipyardConsoleBoundUserInterface.cs
+++ b/Content.Client/_NF/Shipyard/BUI/ShipyardConsoleBoundUserInterface.cs
@@ -6,10 +6,14 @@ using static Robust.Client.UserInterface.Controls.BaseButton;
 using Robust.Client.UserInterface;
 using Content.Client.Shuttles.Save;
 using Robust.Client.UserInterface.Controls;
-using Robust.Client.UserInterface.XAML;
+using System.Text.RegularExpressions;
 using Robust.Shared.IoC;
 using Robust.Shared.Log;
 using System.Linq;
+using Robust.Shared.Serialization.Markdown;
+using System.IO;
+using Robust.Shared.Serialization.Markdown.Mapping;
+using Robust.Shared.Serialization.Markdown.Value;
 
 namespace Content.Client._NF.Shipyard.BUI;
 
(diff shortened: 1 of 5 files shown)


Content.Client/_NF/Shipyard/BUI/ShipyardConsoleBoundUserInterface.cs
```diff
@@ -6,16 +6,24 @@
 using Robust.Client.UserInterface;
 using Content.Client.Shuttles.Save;
 using Robust.Client.UserInterface.Controls;
-using Robust.Client.UserInterface.XAML;
+using System.Text.RegularExpressions;
```
TheRealMasterChief117:
```suggestion
```

Content.Client/_NF/Shipyard/BUI/ShipyardConsoleBoundUserInterface.cs
```diff
@@ -130,12 +146,33 @@ private async void OnLoadShipButtonPressed(BaseButton.ButtonEventArgs args)
         }
     }
 
-    private void OnSavedShipSelected(ItemList.ItemListSelectedEventArgs args)
+    private async void OnSavedShipSelected(ItemList.ItemListSelectedEventArgs args)
```
TheRealMasterChief117:
```suggestion
    private void OnSavedShipSelected(ItemList.ItemListSelectedEventArgs args)
```

TheRealMasterChief117:
Shameless

PR #223, opened 2026-06-14
Revert "Fix/Server Exception Sweep #2"

DDrakov:
Reverts Triad-Sector/Triad_Sector#220

Need some more testing and causing problems

**Changelog**
:cl:
- fix: Revert the #220 PR because issues


rebaserHEAD:
My shame is eternal and mercy is not deserved. I die by the sword for this indiscretion.

PR #236, opened 2026-06-19
Fix/Expedition FTL Cooldown Guarding

rebaserHEAD:
## About the PR
Fixes a data-loss bug where ending a salvage expedition early, before the ship's FTL drive finished its arrival cooldown, deleted the ship and everyone aboard it. The early-return is now blocked until the drive is ready, and the expedition map can no longer be torn down while a ship is still on it.

## Why / Balance
When a ship FTLs into an expedition it stays in its arrival cooldown (mass-scaled, up to ~50s) and keeps its `FTLComponent` the whole time. The expedition cleanup auto-FTLs shuttles back out, but it skips any grid that still has an `FTLComponent`, then `QueueDel`s the map, deleting the un-evacuated ship with it. Returning early lands you right inside that window. The return lock reuses the same `HasComp<FTLComponent>` cooldown check the mission-claim path already uses, and the map-teardown guard lets the retrying auto-FTL evacuate a stranded ship instead of deleting it. This is a bug fix, not a balance change.

## Media
{F614E74E-A0DC-49AB-A7A6-844CED5246BD}

## Requirements
- [X] I have read relevant guidelines/documentation to this PR found on our devwiki.

## How to test
1. Buy/load an expedition vessel (e.g. a Duran), board it, and fuel it.
2. From the salvage expedition console, claim and FTL to an expedition.
3. The moment you arrive, hit the early-return button.
   - Before: when the timer expires, the ship and everyone aboard are deleted.
   - After: the console denies with "The FTL drive is still spooling down from the jump in. Wait for it to cool down before returning." until the drive's cooldown ends; once it's ready, returning evacuates the ship normally.
4. Let an expedition also run to completion to confirm normal end-of-expedition returns are unaffected.

## Breaking changes
None.

**Changelog**
:cl:
- fix: Returning from a salvage expedition early no longer deletes your ship if you do it right after jumping in.

### DIFF
--- a/Content.Server/Salvage/SalvageSystem.Runner.cs
+++ b/Content.Server/Salvage/SalvageSystem.Runner.cs
@@ -160,6 +160,20 @@ public sealed partial class SalvageSystem
         QueueDel(ev.FromMapUid.Value);
     }
 
+    // Triad: true while any player ship (ShuttleComponent grid) is still parented to the expedition
+    // map. Dungeon/asteroid terrain grids aren't shuttles, so they don't block teardown.
+    private bool ExpeditionMapHasShuttle(EntityUid map)
+    {
+        var query = AllEntityQuery<ShuttleComponent, TransformComponent>();
+        while (query.MoveNext(out _, out var xform))
+        {
+            if (xform.MapUid == map)
+                return true;
+        }
+
+        return false;
+    }
+
     // Runs the expedition
     private void UpdateRunner()
     {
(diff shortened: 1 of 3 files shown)


Content.Server/Salvage/SalvageSystem.Runner.cs
```diff
@@ -269,7 +283,12 @@ private void UpdateRunner()
 
             if (remaining < TimeSpan.Zero)
             {
-                QueueDel(uid);
+                // Triad: don't tear down the expedition map while a player ship is still on it.
+                // The auto-FTL above retries every tick, so a ship stranded by its FTL-drive
+                // cooldown is evacuated once the drive is ready; deferring until the map is clear
+                // turns "ship + crew deleted" into "expedition ends a few seconds late".
+                if (!ExpeditionMapHasShuttle(uid))
+                    QueueDel(uid);
```
DDrakov, line 291: In the event of exped failing and doesn't delete the ship, would it try to delete the exped back again?

DDrakov (approved):

Content.Server/Salvage/SalvageSystem.Runner.cs
rebaserHEAD: Discussed in Maintainer chat. Approved due to the inability to reach this bad state anymore. User can no longer put the exped in to QueueDel inappropriately.

PR #249, opened 2026-06-22
Cherrypick: Adds the Ore Magnet (#3921)

mnva0:
## About the PR
Ports https://github.com/Monolith-Station/Monolith/pull/3921, which was merged recently
- 10k spesos at a FlatpackVend
- Under "Salvage Equipment" research, which also has industrial ore processors
- Rare scrap loot I think?
- Picks up ore within 30 meters when enabled. You can get a rough gauge of the range if your shuttle console is *very* close to it as it'd be around the 32m ring

Note that this uses power but doesn't actually need it
Construction can take *any* capacitor, interestingly
This has the same sprite as the salvage magnet

- Tested to work fine with ship saving with the magnet on
- Flatpack works
- Researched recipe works
- Construction works

## Why / Balance
Massive convenience for mining and possibly a reason to actually set up and make temporary infrastructure on larger mining sites, which would be cool. Mining will be far faster for smaller asteroids

I'm not entirely sure if we **want** this given the fact we're persistent? Monolith is balanced around nonpersistence and acquiring ore quickly is probably an actual consideration?
Unsure if we want to encourage mining being ridiculously quick either? This wouldn't *fully* replace actual crew I think, but it'd be very close
You can delete ore if you deconstruct it. The ore still has to be slooowly dumped out and hauled back

I think I'd also like @rebaserHEAD's opinion on the performance implications of this as that's something I'm unfamiliar with. I kind of just want feedback in general on if this should be ported
- Should the range be debuffed at all?
- Should I try to add a power requirement?
- Should I add a radar blip? (so it's *very* obvious where your ore vanishes to)
- Should I make it unanchorable? Currently, it's just a heavy long-ranged ore box. Making it unanchorable would mean people can't haul around the huge industrial magnet like an ore box. And—because it voids ore when deconstructed—it would encourage setting up little waystations for it on large mining sites
- Should I make dumping it out slower?

## Media
image

## Requirements

## How to test
1. Go through the list of what I tested above

**Changelog**
:cl: Monolith
- add: The ore magnet has been ported from Monolith. It can be bought for 10k credits at a FlatpackVend or found under the Salvage Equipment research. It grabs all ore within 30 meters. Deconstructing it will delete the ore it contains.

### DIFF
--- a/Resources/Prototypes/Recipes/Lathes/Packs/cargo.yml
+++ b/Resources/Prototypes/Recipes/Lathes/Packs/cargo.yml
@@ -27,6 +27,7 @@
   id: CargoBoards
   recipes:
   - OreProcessorIndustrialMachineCircuitboard
+  - OreMagnetMachineCircuitboard # Mono
   # - CargoTelepadMachineCircuitboard # Frontier
   - ShuttleGunKineticCircuitboard
   - WeaponTurretM25Circuitboard # Mono
--- a/Resources/Prototypes/Research/industrial.yml
+++ b/Resources/Prototypes/Research/industrial.yml
@@ -12,6 +12,7 @@
   - WeaponGrapplingGun # Frontier
   - MineralScannerEmpty
   - OreProcessorIndustrialMachineCircuitboard
+  - OreMagnetMachineCircuitboard
   - ClothingMaskWeldingGas
   - ClothingShoesBootsJump
   - MechEquipmentDrill # Goobstation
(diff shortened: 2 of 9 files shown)


rebaserHEAD:
Load from this is relatively low. Magnet system is the least of our worries, to my mind. Let me look it over and I'll let you know tonight!

rebaserHEAD:
Should the range be debuffed at all?
- 32 should be fine. we can tune what we see in use in the wild.

Should I try to add a power requirement?
- This probably should have SOME form of power requirement since it's an electro-magnet.

Should I add a radar blip? (so it's very obvious where your ore vanishes to)
- EH. No opinion here. Either or is fine in my book? If you built it you should know where you put it, it's supposed to be a static object.

Should I make it unanchorable? Currently, it's just a heavy long-ranged ore box. Making it unanchorable would mean people can't haul around the huge industrial magnet like an ore box. 
- Nope, It should require power.

Should I make dumping it out slower?
- Nah, but it WOULD be cool if it could dump in a spot in front or behind it so this could be used to feed a belt. Maybe add a link to a purge function along with an on/off toggle? Just a thought!

TschudyNF:
Maybe the range could scale based on the installed capacitors?

mnva0:
I am closing this PR and will instead keep this as an entry on my to-do list until I get around to it

PR #253, opened 2026-06-23
Perf/Skip Dead IsWeightlessEvent Raise for Inventory-Less Bodies

rebaserHEAD:
## About the PR
Guards the `IsWeightless` event raise in `SharedGravitySystem`. The method raised an `IInventoryRelayEvent` (`IsWeightlessEvent`) for every moving body every tick from the physics prestep loop, even for entities with no handler for it. The raw event has exactly three direct subscribers (`InventoryComponent` relay, `MagbootsComponent`, `NFMoonBootsComponent`); when an entity has none of those, the raise dispatches to nothing, so we skip it and fall through to the existing grid/map gravity check directly. Note for maintainers: any future component subscribing to the raw `IsWeightlessEvent` must be added to the guard, which an inline comment calls out.

## Why / Balance
For inventory-less mobs (carp, wildlife, projectile debris) the raise paid a directed event-bus lookup per body per tick for zero handlers; dotnet-trace measured it at ~9% of `PhysicsSystem.Update` under a mob-heavy load. Skipping it is bit-equivalent to raising it with no handler: the event struct stays `Handled=false` and the method falls through to the same gravity check. This is a guard, not a cache, so nothing goes stale and there is no prediction rubberbanding risk. No intended gameplay change.

## Media
N/A, backend performance change with no intended visible behavior change.

## Requirements

## How to test
- Confirm an inventory-less mob (e.g. a carp) is weightless in space and not weightless on a gravity-enabled grid, same as before.
- Confirm a crewmember is weightless in space and not weightless under gravity.
- Confirm magboots and moonboots still keep their wearer grounded in zero-g.

## Breaking changes
None.

**Changelog**
:cl:
- tweak: Optimized weightlessness checks for better performance with many creatures.

### DIFF
--- a/Content.Shared/Gravity/SharedGravitySystem.cs
+++ b/Content.Shared/Gravity/SharedGravitySystem.cs
@@ -1,4 +1,6 @@
+using Content.Shared._NF.Clothing.Components; // Triad
 using Content.Shared.Alert;
+using Content.Shared.Clothing; // Triad
 using Content.Shared.Inventory;
 using Content.Shared.Movement.Components;
 using Robust.Shared.GameStates;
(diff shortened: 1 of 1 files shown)


DDrakov (approved):
Shouldn't improve much but it's a better thing when lots of entities are present

PR #256, opened 2026-06-23
Increases Turret Range and Add Faction Turrets (TDF, SD, TIC & Syndicate)

Wuffindy:
## About the PR

Increases the base default range of all turrets to 12, and added 2 more variation that could be used for future dungeon and mapping endeavours,

**Default 10 -> 12 Tiles
MediumRange 16 Tiles
LongRange 20 Tiles**

current turrets that gets a ranged variants are only the TDF and the Syndicates, with Syndicate lacking in Heavy Ballistic for now, since those packs too much punch, should mapping wants to make more of that i'll add in another PR
## Why / Balance

Increases the default range because players can always peek/zoom over the turret which is cheesy, and the turret can only stand there watching, and taking all the damage, keep in mind even with 12 tiles, while zoomed they are still slightly visible before, they shoot you for default variants
Addition of turrets are for mapping and event purposes, previously requested by the admins, and also addition personally for future dungeon mapping
## Media

image
Visual range of each turret range, 12-16-20

default changes to 12, which is about the max of your zoom, the moff is standing safely, not attacked at the image, if they move slightly more, they will be shot
image

Syndicate Turrets
image
image
no Heavy Ballistics for syndicates for now

TDF Turrets
image
image

SD Turrets:
media to be added

TIC Turrets:
media to be added

## Requirements

## How to test

spawn a turret
get in range of the turret
bullet rain upon ye

## Breaking changes

none so far but, i have added HideSpawnMenu on some turrets, especially the vanilla ones, that way if anyone were to make dungeons, it is recommended to use the Triad Turret

**Changelog**

🆑
- add: Added more turrets for TDF, Future Factions and Syndicate
- tweak: Changed the laser type of TDF Laser turret into something stronger
- tweak: All default turret range has been increased to 12 tiles, about the same range as your peek/zoom view


DDrakov:
> So if im reading this right, turrets are basically being given a 50% damage buff as well as better range? They already lock on from nearly a full screen away as it is.

Press Space. It'll increase your view significantly more than most turret does.

TschudyNF:
> > So if im reading this right, turrets are basically being given a 50% damage buff as well as better range? They already lock on from nearly a full screen away as it is.
> 
> Press Space. It'll increase your view significantly more than most turret does.

Current state, yes.  Might not be the case with the upgraded ones being pushed here.

Wuffindy:
i dont know how you even read the 50% increase in damage stop exaggerating and spreading misinformation for your third time, also press space, currently players can outview the range of any NPC (that will be adjusted in a different PR, it will come.)

do you have any idea how many cheese is in PVE content

TheRealMasterChief117 (approved):

PR #261, opened 2026-06-25
Rebalance Gas Deposits

Dimasw99:
## About the PR
Add a unique deposit type for each asteroid to be used instead of gas type deposits.
Increase amount of moles you get per asteroid to 45k Moles on average considering scanning for failed asteroids.

## Why / Balance
Asteroid mining is a pain. It takes both a ton of time trying to find the correct gas and when you do find a deposit they can sometimes just be a 10k moles or less. By also creating a deposit type for each asteroid it allows for much better fine tuning for what gasses and ratios you want for each asteroid. 

Current numbers might be also generous though considering the time it takes to scan, mine, setup, harvest and selling it shouldn't be too bad for the already crying economy of Triad. I expect this to be tweaked a few times to get this right but so far I consider this a good start in trying to make this gameplay style worthwhile.

## Media
grafik
A quick glance at the total value, moles, average moles of each type per deposit and total moles compared to everything else.

**Changelog**

:cl:
- tweak: Asteroids now have their own unique gas deposits with specific gas ratios for each type.
- tweak: Significantly increased the amount of moles per asteroid on average.

### DIFF
--- a/Resources/Prototypes/_NF/Entities/World/Debris/asteroids.yml
+++ b/Resources/Prototypes/_NF/Entities/World/Debris/asteroids.yml
@@ -32,37 +32,24 @@
     - type: RandomEntityPopulator
       entries:
       - params:
-          prob: 0.5
-          min: 1
-          max: 2
+          prob: 0.4
+          min: 2
+          max: 6
           canBeAirSealed: true
         entries:
-        - id: GasDepositOxygen
+        #Triad Change start
+        - id: GasDepositBaseAsteroidVeryLarge
           orGroup: deposit
-        - id: GasDepositNitrogen
+        - id: GasDepositBaseAsteroidLarge
           orGroup: deposit
-        - id: GasDepositCarbonDioxide
+        - id: GasDepositBaseAsteroid
           orGroup: deposit
-          # Mono Changes Start
-        - id: GasDepositPlasmaVerySmall
+        - id: GasDepositBaseAsteroidSmall
           orGroup: deposit
-        - id: GasDepositPlasmaSmall
+        - id: GasDepositBaseAsteroidVerySmall
           orGroup: deposit
-        - id: GasDepositPlasma
-          orGroup: deposit
-        - id: GasDepositPlasmaLarge
-          orGroup: deposit
-        - id: GasDepositPlasmaVeryLarge
-          orGroup: deposit
-        - id: GasDepositMuddlePlasma
-          orGroup: deposit
-          # Mono Changes End
-        - id: GasDepositAir
-          orGroup: deposit
-        - id: GasDepositMuddleEven
-          orGroup: deposit
-        - id: GasDepositMuddlePlasma
-          orGroup: deposit
+        #Triad change stop
+
 # endregion Rock Table
 
 # region Snow Table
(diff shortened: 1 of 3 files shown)


Wuffindy:
okay so i don't fully understand yet on what are the percentage the chances of Gas deposits in asteroid works but, assuming by the current changes, that a whole majority of bloated individual that got removed there which is understandable, and replacing with a simpler combined option, the chances of getting every other type becomes a 20% no matter the size

however you also changed the total min/max to 2 to 6 which is far more than previous ones, considering that you now have a bunch more of everything instead of individual ones

which is like 7x more than the original from any of the veins since you're guaranteed of getting all types of gas from any vein

with my current lack of understanding regarding atmos stuff, i would suggest toning it to a more reasonable of 2-4 veins (1 is unfairly rare, 6 feels like it's too much with everything at a flat 20% chance of getting any type, no matter the size)

but that's just my take, as a non atmos player perspective, with only basic knowledge on shuttle piping. though i'm interested to see and understand how this would play out though, i hope it's better than what we had previously cause i heard that asteroid gas mining is ass before

Dimasw99:
If you look at the spreadsheet I posted in media it shows you the average mole target for each gas and asteroid type for their unique deposit. I then simply took 50%, applied it to high and low to get my variance to make things look more interesting and natural.

I would also like to know how you got your 7x

Here is a screenshot detailing the value and mole count of each old asteroid:
grafik
And here is a screenshot of every new deposit type with its average value and mole count.
grafik

Both value and mole count has been on average buffed but as you can see it is far away from the 7x.
If you were to take the average of every single asteroids value then you would get 27312.6 spesos per old asteroid and 36088 for the new ones. It is only a 32% buff across the board

Wuffindy:
the 7x is just a quick assumption based on what i'm seeing with what knowledge that i have at the moment, it's definitely not an accurate number to represent the changes that you've done.
the thought for it was that,  i'm assuming for min 1 and max 1 type of older asteroid version, you would get 1 type of gas from the vein (ex: Plasma L) and then you wont get other types of gas, unless it's a muddled one

which with the new one, you are pretty much guaranteed to get all types of gas from any vein no? just that the total mol is different based on the types of roids you're mining on that's the, that's where i didn't account the total mol and assumed the 7x for simplification on my end, my math was very off there so my apologies for the misunderstanding.

suppose the charts you sent there is more refined and more understandable if using the average of it, and you know what that's fair, with what you mention on the average comparison between old and new asteroid

TheRealMasterChief117 (approved):
New numbers look good, I also like the different asteroids having differing gas-heavies, makes it so if you're after a certain gas you can look for one

This info should be communicated in the guidebook, but it's fine as-is for now

Dimasw99:
> This info should be communicated in the guidebook, but it's fine as-is for now

I don't have much guide book experience but can you upload images? If yes. I might try creating a quick and dirty graph

TheRealMasterChief117:
> > This info should be communicated in the guidebook, but it's fine as-is for now
> 
> I don't have much guide book experience but can you upload images? If yes. I might try creating a quick and dirty graph

You can upload it as a sprite, then make it an entity and attach that entity
Sort of a hacky way, but works

You can also just make the table in the guidebook too, fairly simple with table elements
image

TheRealMasterChief117:
Should list the asteroid types and the gases within if you do this

Dimasw99:
grafik
Quick and dirty example made in open office. I might try to get it in a prettier version in a different program

DDrakov (approved):
Goida

PR #287, opened 2026-07-06
Cryoxadone Revertion

Wuffindy:
## About the PR

in Mono, Cryoxadene is a one med for all, obsoleting every other type of cryogenic medicines, such as Doxa, Stellox, and etc

This PR removes all non airloss type of healing from cryoxadone, and now it heals -8 Airloss instead
## Why / Balance

Cryox should not have been a one med for all, also this was talked about in discord
## Requirements

## How to test

Check Reagent Guidebook of Cryoaxdone
Test out Cryox in a tube, with burn/brute/poison and airloss, should only reduce the airloss now
## Breaking changes

should be none
**Changelog**

🆑 
- tweak: Cryoxadone has been reverted to Airloss only Cryogenic medicine

### DIFF
--- a/Resources/Prototypes/Reagents/medicine.yml
+++ b/Resources/Prototypes/Reagents/medicine.yml
@@ -204,10 +204,7 @@
             max: 213.0
           damage:
           # todo scale with temp like SS13
-            Airloss: -3
-            Brute: -2
-            Burn: -2
-            Toxin: -2
+            Airloss: -8 # Triad
         - !type:GenericStatusEffect # Mono edit
           conditions:
           - !type:OrganType #
--- a/Resources/Prototypes/Reagents/medicine.yml
+++ b/Resources/Prototypes/Reagents/medicine.yml
@@ -204,7 +204,10 @@
             max: 213.0
           damage:
           # todo scale with temp like SS13
-            Airloss: -8 # Triad, 8 Airloss instead of 10
+            Airloss: -3
+            Brute: -2
+            Burn: -2
+            Toxin: -2
         - !type:GenericStatusEffect # Mono edit
           conditions:
           - !type:OrganType #
--- a/Resources/Prototypes/Reagents/medicine.yml
+++ b/Resources/Prototypes/Reagents/medicine.yml
@@ -204,10 +204,7 @@
             max: 213.0
           damage:
           # todo scale with temp like SS13
-            Airloss: -3
-            Brute: -2
-            Burn: -2
-            Toxin: -2
+            Airloss: -8 # Triad, 8 Airloss instead of 10
         - !type:GenericStatusEffect # Mono edit
           conditions:
           - !type:OrganType #


Dimasw99:
Obsoleting is a very strong word. The other meds have the incredible abillty of healing percentage damage so if you have someone with 2000 burn because he got spaced then the first tick would heal a whopping 40 damage plus its other static values. Cryox current healing is in the spot of enough that it is nice to use but incredibly inefficent when it comes to huge amount of damage

Eillyrea:
+1 
Cryox is too powerful in it's current iteration.  It's the omni of cryo chems

Wuffindy:
> Obsoleting is a very strong word. The other meds have the incredible abillty of healing percentage damage so if you have someone with 2000 burn because he got spaced then the first tick would heal a whopping 40 damage plus its other static values. Cryox current healing is in the spot of enough that it is nice to use but incredibly inefficent when it comes to huge amount of damage

couldn't respond earlier since i was too busy at work, the one you're overlooking at is Necrosol, which was supposed to be the omni of cryo chem, before Cryoxadone decided to curbstomp it to the ground, replacing it while remaining the cheapest variant of all cryo chem. if it was in mono, it is understandable since it's a more combat focused server and they needed to be fast, however in Triad, this over convenience is a detriment on the long term.

If Obsolete isn't the right word, then Necrosol is an absolutely powerful end game fancy milk Cryox with roughly similar healing of cryox in all healing the with a far more expensive recipe that somehow no one uses most of the time.

anyhow specialized are specialized which should always be superior than omni heal for obvious reason, Cryox is just the base and basic of em all, necrosol is made of omni, which should've been the one to do what an current Cryox do

TheRealMasterChief117:
> Obsoleting is a very strong word. The other meds have the incredible abillty of healing percentage damage so if you have someone with 2000 burn because he got spaced then the first tick would heal a whopping 40 damage plus its other static values. Cryox current healing is in the spot of enough that it is nice to use but incredibly inefficent when it comes to huge amount of damage

I can attest to it being horrible at healing high amounts of damage, as I have dealt with trying to get rid of 800 poison with just cryoxadone before

However, cryoxadone still works amazing as an 'omnizine' for people with spread out damaged types
It serving as a precursor is so you can get the specific cryogenic chems would be better here

TschudyNF:
"Cryox should not have been a one med for all, also this was talked about in discord" is extremely poor reasoning for such a significant change.  Can you provide a link to the discussion where it was "talked about"?

Wuffindy:
> "Cryox should not have been a one med for all, also this was talked about in discord" is extremely poor reasoning for such a significant change.  Can you provide a link to the discussion where it was "talked about"?

https://discord.com/channels/1421044907157618751/1501129123462906008
This was just the *Public* thread discussion

More reason on the earlier comments above, go check that out

DDrakov (approved):
Approved, cryoxadone ingredients is not all meds combined into one. the ingredient itself is Dex, Oxy and Water itself idk why it's all curing where it shouldn't be. If we want all cure cryo med like Omni, we should make a unique one and makes it more complicated.

PR #306, opened 2026-07-09
Adds the Basic Medal Case to Loadout

mnva0:
## About the PR
This PR adds the medal case to the loadout under Trinkets (for a price)

## Why / Balance
This is a kind of tricky one. This will allow captains to give medals to others or their crews, which would make for very cool roleplay. This WOULD devalue medals a bit due to accessibility, and I'm not certain if this would be worth it

The TDF (and likely thus the Overseer) would have still access to their special medals, but this would mean the other medals are far less valuable due to being easily obtainable. Being given a regular medal from an Overseer wouldn't be as big of a situation because *anyone* could. Or you could give yourself a medal. But ideally a captain would be giving you a medal for doing something cool
Having a medal that is "signed" by the Overseer might have *some* weight to it? Maybe I could rework medals to have an *actual* component to be used as paper so signatures aren't forgeable?
Medals *could* also be made non-persistent, but that would make it virtually impossible to use them as collectors items or something to reminisce on. Net negative

Something to consider is that if someone pays the price, they're going to want to spend all the medals. If the price is increased to a higher number, it's likely that the players will still spend all the medals they can

## Media
image
image

## Requirements

## How to test
1. Open loadout. Most jobs
2. Go to Trinkets
3. There is a medal case

**Changelog**

:cl: Minerva
- add: The medal case has been added to loadout. Captains can now give a small variety of medals to crewmembers who go above and beyond in their duties.

### DIFF
--- a/Resources/Prototypes/_NF/Loadouts/contractor_loadout_groups.yml
+++ b/Resources/Prototypes/_NF/Loadouts/contractor_loadout_groups.yml
@@ -900,6 +900,7 @@
   - ClothingNeckTransPin
   - ClothingNeckAutismPin
   - ClothingNeckGoldAutismPin
+  - MedalCase # Triad
 
 - type: loadoutGroup # Mono - removed standard supply boxes.
   id: ContractorBoxSurvival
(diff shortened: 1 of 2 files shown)


DDrakov (approved):
good for larp

PR #353, opened 2026-07-14
Omega, Station, Enforced Max Console Server Saving Contraband

Wuffindy:
## About the PR

before anyone try to spread this PR around again without understanding, this only affects Omega, Station, and Enforced Max Console Server, which does not exist nor obtainable at any way whatsoever, even so obtaining it will be removed on the administrative side, unless it's something that has been discussed

also seperating this from the UGCS pr, which is on hold
## Why / Balance

there is no balancing for these items, should only for limited events and etc
## Requirements

## Breaking changes

should be none

**Changelog**

not really player facing? as it it is unobtainable to begin with
- tweak: Added Saving Contraband for Omega, Station & Enforced Max Console Gunnery Server


DDrakov (approved):
Approved as per discussion

PR #361, opened 2026-07-14
Finally Fix Station Records Computers with Ships

TheRealMasterChief117:
## About the PR
``StationRecordsFilter`` was causing serialization issues, so we reset the consoles when a ship is saved
This wouldn't throw if a player didn't mess with their station records computers, it's still good to fix this

## Technical details
Added ``SetActiveKey`` and ``SetFilter`` static methods to general records system so you can set that data

**Changelog**
:cl:
- fix: Fixed station records computer filters causing your ship to be unable to be saved.


TheRealMasterChief117:
(self merges)

PR #381, opened 2026-07-19
Fix: Repair Humanoid Sprite-Layer Drift and Add Dual-Layer Wings

rebaserHEAD:
## About the PR
IPCs never shared the canonical humanoid layer list, so their `Sprite.layers` order had drifted from every other species and the `Wings` layer was never actually mounted. This resyncs the divergent layers and gives IPC wings a proper behind/over layer pair so they render correctly from every facing.

## Why / Balance
Wing markings on IPCs drew on top of everything (arms, backpack, the lot) because the `Wings` layer was never in the sprite's layer list, so the humanoid system appended it last. The fix mirrors the tail system: a low `Wings` layer stays behind the body for the front and side facings, and a new `WingsOversuit` layer draws over the body for the back facing. The marking's `layering`/`colorLinks` route each directional frame to the right layer, and `colorLinks` keeps the over-frame colored like its base. No wing art was redrawn; the back-facing frame is re-derived onto the over layer. While in there I resynced the other drifted layers against `BaseMobSpecies` (ID/neck order, missing balaclava/helmet-cover/helmet-attachment bookmarks) and commented out the dead digitigrade stencil layers. No balance impact, cosmetic only.

Please note: the display on the character creation screen is still goofed up on this but I can't be bothered at this point to untangle that disaster as well.

## Media
image
Front - No clipping!
image
Back - Also no Clipping!

## Requirements

## How to test
1. Spawn as / spawn an IPC.
2. In character setup add the Robotic or Merciful wing marking.
3. Rotate through all four facings: wings sit behind the body from the front and sides, and over the body from the back.
4. Equip a balaclava, a helmet cover, and a helmet attachment and confirm they render.
5. Give the IPC an ID card and a neck-slot item and confirm draw order matches other species.

## Breaking changes
Adds a new `WingsOversuit` value to `HumanoidVisualLayers` and its `MarkingCategories` mapping. Additive and appended at the end of the enum (safe for net serialization), no renames or removals.

**Changelog**
:cl:
- fix: IPC wings now render in the correct layer from every direction instead of drawing over everything.
- fix: Balaclavas, helmet covers, and helmet attachments now render on IPCs.
- fix: Fixed the draw order of IPC ID cards and neck-slot items.


TheRealMasterChief117 (approved):
Good for now. Would be good to get Wizden's per-direction layering system in, would fix all markings similar to this eventually (if that's even merged on wizden)

PR #408, opened 2026-07-23
Privacy Holoprojectors

TheRealMasterChief117:
## About the PR
Adds the privacy holoprojectors. You can change the description of the holosign on the projector, giving more info about what is going on past the private sign, the option to make it explicit or not (hidden from people with the consent setting off) is available.

They are available in the autolathe and service techfab, as well as loadouts under the fun category

## Media

https://github.com/user-attachments/assets/ea0c98e2-5bbf-4251-97cb-5f99ee230b3f

## Technical details
Adds the ``ChargeHolosignPlacedEvent`` which is used when a holosign is placed so the label system can intercept it and label it correctly

**Changelog**
:cl:
- add: Added privacy holoprojectors, available in loadouts under the fun category and at an autolathe and service techfab. You can press Z to access its interface, where you'll be able to give a description of the events happening past the privacy holoprojectors.


DDrakov (approved):
The UI can be improved but that's future issue. Feel free to merge

PR #434, opened 2026-07-27
Fix Sleeping Drone Brains Piloting Across the Sector After a Target F…

rebaserHEAD:
## About the PR
Sleeping an NPC only removes `ActiveNPCComponent`; it never tears the HTN plan down, and ship steering runs off the mover rather than the NPC update loop. So a drone that fell asleep mid-chase kept its `ShipSteererComponent` alive, quietly piloting toward blackboard coordinates anchored to a ship that had FTL'd away, then following it across the whole sector once it dropped out of FTL and waking up hot wherever it landed.

This adds `ShipNpcSleepSystem`, which tears ship brains down when they sleep: shuts the plan down, scrubs the `Target`/`TargetCoordinates` keys, and stops steering and targeting. `ShipSteeringSystem.Stop` now also restores inertia dampening for NPC cores, since steering leaves dampening Off while thrusting; a torn-down hull now brakes and parks instead of coasting ballistic at combat speed, which puts it in reach of the normal grid cleanup cycle. Autopilot consoles are excluded from both paths and keep the pilot's dampening selection.

## Why / Balance
FTLing away from a drone fight should actually end the fight. Before this, drones from the outer rings could end up in the core regions chasing their escaped target, then re-acquire whoever was nearest. Side effect worth knowing: a depowered or EMP'd drone now brakes to a stop instead of drifting off, which reads as "disabled" rather than "runaway".

## Media
N/A, server-side fix.

## Requirements

## How to test
Spawn a combat drone (a rammer or `quake`), aggro it with a ship, then FTL away with nobody else nearby. VV the drone's AI core: `ActiveNPCComponent` and `ShipSteererComponent` should both be gone, the drone grid's dampening should read Dampen with velocity bleeding off, and `GridCleanupGridComponent`'s accumulator should start climbing. It should not follow you to your destination. Builds clean and the teardown paths are verifiable via VV; this still wants a live two-ship shakedown before merge.

## Breaking changes
None.

**Changelog**
:cl:
- fix: Combat drones no longer chase ships across the sector after they FTL away, and disabled drones brake to a stop instead of drifting forever.

### DIFF
--- a/Content.Server/_Mono/NPC/HTN/ShipSteeringSystem.cs
+++ b/Content.Server/_Mono/NPC/HTN/ShipSteeringSystem.cs
@@ -752,6 +752,20 @@ public sealed partial class ShipSteeringSystem : EntitySystem
             return;
 
         RemComp<ShipSteererComponent>(ent);
+
+        // Triad: steering leaves dampening Off while thrusting, so a teardown mid-flight
+        // (target FTL'd away, brain slept, operator failed) left NPC hulls ballistic at combat
+        // speed, migrating across the sector. Restore Dampen so they bleed off velocity and
+        // park for grid cleanup. Consoles keep the pilot's Drive/Cruise/Park selection (#4064).
+        if (TerminatingOrDeleted(ent)
+            || _consoleQuery.HasComp(ent)
+            || Transform(ent).GridUid is not { } grid
+            || TerminatingOrDeleted(grid)
+            || !_physQuery.TryComp(grid, out var body)
+            || !_shuttleQuery.TryComp(grid, out var shuttle))
+            return;
+
+        _shuttle.SetInertiaDampening(grid, body, shuttle, Transform(grid), InertiaDampeningMode.Dampen);
     }
 
     private ref struct SteeringContext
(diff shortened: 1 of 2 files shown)


TheRealMasterChief117 (approved):

TheRealMasterChief117:
Instead of dampening, should it go into anchoring mode instead when players FTL away? In my testing with a quake, the drone seems to move a bit as if they were in cruise mode.

DDrakov:
> Instead of dampening, should it go into anchoring mode instead when players FTL away? In my testing with a quake, the drone seems to move a bit as if they were in cruise mode.

I think that Mono removed dampener when they're making the autopilot, so grid will just flings off into distances and not stopping because of it.

TheRealMasterChief117:
So, it sets it into Dampening. Which is equivalent to Drive on the shuttle console, so it'll stop eventually, but if the drone was going super fast, it'll go into space until friction stops it naturally. Setting it to Anchor (park mode) would be better so it stops in place

DDrakov:
True, I think it's fine for now probably gonna tweak about it later.

PR #446, opened 2026-08-01
New Expedition Ship: Behir

gaywolfgirl6:
## About the PR
Added a new DIRE Ship: The Behir, and made a pretty Guidebook for it :3

## Why / Balance
A new fun Expedition ship, to help add flavoring to the server. 

## Media
image

image

image
image

## Requirements

## How to test
Purchase the ship, fly it around, look at the guidebook! I'm really proud of it.

## Breaking changes
New Expedition Ship: Behir, Guidebook for it

**Changelog**

:cl:
- add: Added new Expedition ship: the DIRE Behir.
- add: Added Guidebook for the DIRE Behir.


Eillyrea:
One of the external airlocks at the front is not glass, two are, unsure if intentional
Not sure about putting 24 conveyor belts in a crate, it's not the worst but is quite a lot of them xD
Bridge light switch has the frame named, not the button
Random brown pipes all over the ship for unknown reasons, looks potentially aesthetic but I'm uncertain 
Two gunnery consoles on the bridge
I got it's for aesthetics, but please keep power systems so they're understandable to people trying to modify them 
image

This is confusing to new people who don't know engineering who may be trying to modify the ship

Scrubber and vent under the z-pinch are a bit unnecessary, and aren't linked to an air alarm

image

Ship is gorgeous

gaywolfgirl6:
> Ship is gorgeous

thank you thank you thank 7ou thank you!!!!!!! awroof wuff :3 i can tone down on the HV cabling under the majority of the ship. the HV Cabling, and MV Cabling, runs along nose to tail, so as to remove the need to place it from one end to another when the owner of a Behir decides to modify it, extending it from any direction.

The HV cabling is especially simple for those new to engineering to have a direct way to anchor RTG's to it anywhere aboard the vessel, and the presences of such makes installing a another Substation take less than a minute.

The brown pipes underneath each doorway are pre-placed air mixing pipes, should the owners be, for instance, a Hydrakin, and want a unique cooling system that runs through the ship, having them easily modifyable as a base, and not requiring removing tiles underneath doorways to add such. The Air Mix pipe extends to Catwalked flooring for visibility. This can also be helpful if you, say, remove the front center docking door in favor of a Gas Mining thingy.

The scrubber underneath the Z-Pinch is to help deal with the exhaust of the machine, as it produces Carbon Dioxide while active.

The two Gunnery consoles on the bridge are in case of having a dedicated pilot and gunner, which is a very rare edgecase. i was mostly going for rp aesthetics when adding such, but it can be removed if desired.

I averaged the 24 conveyer beltings by doing dorky in character interviews about people that had modified the Behir on other servers it exists, and found that it is, on average, the adequate amount needed to do more than just a loop for their processors and silos, and is around the amount placed on cargo distribution vessels similar to the Behirs size.

i cannot believe i missed the extra pipe i placed in the tail end omg.

Eillyrea:
The console is like 4 tiles from the other gunnery console xD They're basically right next to each other.

Well reasoned responses for everything.  I grrrrrrr at your style but it's clearly got a purpose to everything required.  

Fix the pipe and remove one of the consoles and we can merge it
Grrrrrr

gaywolfgirl6:
MA'AM YES MA'AM

PR #454, opened 2026-08-02
TDF Contraband Permit Console

TheRealMasterChief117:
## About the PR
Adds the contraband permit manager, the long awaited feature that allows for persistent permits and an easy way to manage them for TDF, increasing merc-TDF interaction!

You print out a chip from the console, which allows you to scan a valid permittable contraband item. Then, the permit owner scans their identity on the card by interacting with it. From there, you insert it into the console, go into the grant permit tab. (Only the Chief Enforcer can grant permits!) and enter in a valid reason. then press the grant button.

It also allows you to revoke permits remotely. When a permit is granted or revoked, the permit owner will receive a PDA message.

## Extra
Should be mapped to Auric in reception, briefing, and the CE's office
SOP entry should be added too
Do extensive testoing

## Media
image

## How to test
Print a chip, spawn in an entity with contraband permittable (auth disruptor or something), scan it, then scan your player. Give your character the ``ContrabandPermitGranterComponent`` so they can grant in permits. You need PTL access or above to revoke permits.

Test if it works with ship saving for things that delete (TDF gear)

## Technical details
Added the HumanoidViewer system, which spawns in a clone of the player's character in a paused map, for things like photographs. The console system uses it to get a picture of the player's character without being a live feed of their sprite. This view is also globally PVS, so clients can see the appearance of the view mobs at ALL times. Checks are in place so that this viewer mob is not recreated when a player respawns with the same character.

Items with valid permits also have global PVS, so that you can view them on the console. I am unsure on how performant/correct this implementation of networking the console this way, however. Would like some extra eyes on that.

## Requirements

**Changelog**
:cl:
- add: Added the TDF contraband permit console, used for granting permits to Triad sector citizens. It prints out permit chips. From there, you scan the item then the permit owner will scan themselves with the chip, which allows you to insert it into the console to grant it. It also allows you revoke permits remotely. Contraband permits will persist on ship save.


TheRealMasterChief117:
Done

Content.Server/_Triad/Humanoid/ClientHumanoidViewerSystem.cs
```diff
@@ -0,0 +1,153 @@
...
+                QueueDel(item);
+            }
+
+            _transform.SetParent(viewerMob.Value, PausedMap.Value);
+
+            var viewerComp = EnsureComp<HumanoidViewerEntityComponent>(viewerMob.Value);
+            viewerComp.Session = ev.Player;
+            Dirty(viewerMob.Value, viewerComp);
+
+            EnsureComp<GlobalPvsComponent>(viewerMob.Value);
```
rebaserHEAD: This spawns a fully-equipped doll clone of every player character at spawn and globally replicates it (plus its equipped gear) to every client, whether or not anyone ever touches a permit console. On a full server that's a lot of permanent PVS and network weight for a portrait, and it also means any client can enumerate every character's name and appearance sector-wide from entity data, which is metagame info we currently don't leak.

The character editor already solves this shape client-side: send the appearance data (or `HumanoidCharacterProfile`) in the BUI state and spawn a local dummy on demand, like the lobby dummies. No server dolls, no global PVS. If that's too big a rework for this PR, the cheap patch is lazy dolls: create on first permit grant for that character, delete on last revoke.

Content.Server/_Triad/Humanoid/ClientHumanoidViewerSystem.cs
```diff
@@ -0,0 +1,153 @@
...
+            if (_humanoidQuery.TryComp(uid, out var humanoid))
+            {
+                if (humanoid.Age == ev.Profile.Age && humanoid.Gender == ev.Profile.Gender)
+                {
+                    viewerMob = uid;
+                    break;
+                }
+            }
+        }
+
```
rebaserHEAD, line 88: Two stale-reuse holes on this match path: `LoadProfile` only runs on creation, so a reused doll keeps its old appearance if the player edited their character between shifts, and this age+gender fallback can bind a *different* character (same session, same age and gender) to the old doll's name and face. If dolls stay, refresh reused ones to the current profile:

```suggestion
        }

        // Reused doll: refresh to the current profile in case the character changed
        if (viewerMob != null)
        {
            _humanoid.LoadProfile(viewerMob.Value, ev.Profile);
            _metaData.SetEntityName(viewerMob.Value, ev.Profile.Name);
        }

```

Content.Shared/_Triad/ContrabandPermit/SharedContrabandPermitSystem.Console.cs
```diff
@@ -0,0 +1,307 @@
...
+        {
+            PlayDenySound(ent, user);
+            ConsolePopup(user, Loc.GetString("contraband-permit-console-popup-error-access-denied"), PopupType.SmallCaution);
+            return;
+        }
+
+        var scannedItem = GetEntity(chipComp.ScannedItem);
+        var scannedPermitCarrier = GetEntity(chipComp.ScannedPermitCarrier);
+
+        if (scannedItem == null || scannedPermitCarrier == null)
```
rebaserHEAD: Crash vector: the chip stores NetEntities, and the scanned item or carrier can be deleted between scan and pressing Grant (item destroyed, carrier gibbed or cryo'd). The null check passes on a dangling uid, then `EnsureComp` / `Name()` below throw server-side.

```suggestion
        if (scannedItem == null || scannedPermitCarrier == null
            || TerminatingOrDeleted(scannedItem.Value) || TerminatingOrDeleted(scannedPermitCarrier.Value))
```

Content.Server/_Triad/ContrabandPermit/ContrabandPermitSystem.cs
```diff
@@ -0,0 +1,319 @@
...
+        }
+
+        // Goodbye
+        RemovePermitRecordToSectorService(permitOwner, permitItem);
+    }
+
+    private void OnPermitItemStartup(Entity<ContrabandPermitItemComponent> ent, ref ComponentStartup args)
+    {
+        // So that permit items outside of PVS range can still be viewed
+        _pvs.AddGlobalOverride(ent.Owner);
```
rebaserHEAD: A global override doesn't just send this item: PVS has to send its transform ancestors too, or clients can't attach it. For a permitted gun in someone's backpack that means the carrying player and their grid get pushed into every client's replication set, permanently, per permit. It also makes permitted items trackable sector-wide from the client. This pairs with the doll architecture note: if the console UI carried data instead of live entities, neither override would be needed.

Content.Server/_Triad/ContrabandPermit/ContrabandPermitSystem.cs
```diff
@@ -0,0 +1,319 @@
...
+
+        if (!exists || permitList == null)
+            permitList = new List<NetEntity>();
+
+        permitList.Add(GetNetEntity(permitItem));
+
+        UpdatePermitConsoles();
+    }
+
+    public void RemovePermitRecordToSectorService(EntityUid permitOwner, EntityUid permitItem)
```
rebaserHEAD: Record-key instability leaks stale entries. Add keys by the doll (via `HumanoidView` on the live owner), but if the owner entity is gone by the time the item terminates (cryo, gib), the `TryComp` here misses and this computes a different key, so the record never comes out of `Records`. The console then lists a permit for an item that no longer exists, forever.

No one-click fix since it needs a field plus both call sites changed together: stamp the computed record key onto `ContrabandPermitItemComponent` in `AddPermitRecordToSectorService`, and remove by that stored key here instead of recomputing it.

Content.Server/_Triad/ContrabandPermit/ContrabandPermitSystem.cs
```diff
@@ -0,0 +1,319 @@
...
+                if (mindComp.CharacterName != comp.PermitOwnerName)
+                {
+                    var message = $"{ToPrettyString(user):player} owns a contraband permit with a different logged name." +
+                        $" (Permit Owner Name: {comp.PermitOwnerName}, Player Name: {mindComp.CharacterName})"
+                        + " Possible abuse of the ship saving system may be at play here.";
+
+                    _chat.SendAdminAlert(message);
+                    _adminLog.Add(LogType.EntitySpawn, LogImpact.Medium, $"{message}");
+                }
+            }
```
rebaserHEAD, line 208: `PermitOwnerName` never gets re-stamped, so a legitimately transferred ship fires this "possible abuse" alert on every single load for the rest of time, which trains admins to ignore it. Stamp the name after the first alert:

```suggestion
                }

                // Re-stamp so a transferred ship only alerts once, not on every future load
                comp.PermitOwnerName = mindComp.CharacterName;
            }
```

Content.Shared/_Triad/ContrabandPermit/ContrabandPermitItemComponent.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// The UID of the permit owner.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public EntityUid? PermitOwner;
+
+    /// <summary>
+    /// The mind of the permit owner. Used for checking if a permitted item should stay or be seized on a saved ship.
+    /// </summary>
+    [DataField]
+    public EntityUid? PermitOwnerMind;
```
rebaserHEAD, line 35: These two uids are session-local (load re-stamps them via `InitializePermitItemsOnGrid`, and the save-side check in `ClearPermitItemsOnGrid` runs on live state before serialization), but `[DataField]` writes them into the ship save YAML where they dangle on load. That's the same stale-uid error-noise class we just cleaned up in the error sweep. Drop them from serialization:

```suggestion
    /// <summary>
    /// The UID of the permit owner. Session-local, re-stamped on ship load; never serialized.
    /// </summary>
    [AutoNetworkedField]
    public EntityUid? PermitOwner;

    /// <summary>
    /// The mind of the permit owner. Used for checking if a permitted item should stay or be seized on a saved ship.
    /// Session-local, re-stamped on ship load; never serialized.
    /// </summary>
    [ViewVariables]
    public EntityUid? PermitOwnerMind;
```

Content.Client/_Triad/ContrabandPermit/ContrabandPermitConsoleWindow.xaml.cs
```diff
@@ -0,0 +1,352 @@
...
+                {
+                    _trackedFocusOwner = null;
+                    UpdateFocus(null, null);
+                }
+                else
+                {
+                    _trackedFocusOwner = owner;
+                    UpdateFocus(owner, null);
+                }
+            };
```
rebaserHEAD, line 203: Stale closure on reused rows: this captures `owner` from when the row was created, but rows get recycled for different owners via `UpdatePermitOwnerAndEntries` when the list shifts after a revoke. Clicking focus on a recycled row then sends the wrong owner. Read the container's current field at press time instead:

```suggestion
            newEntryContainer.FocusButton.OnButtonUp += args =>
            {
                var currentOwner = newEntryContainer.PermitOwner;
                if (_trackedFocusOwner == currentOwner)
                {
                    _trackedFocusOwner = null;
                    UpdateFocus(null, null);
                }
                else
                {
                    _trackedFocusOwner = currentOwner;
                    UpdateFocus(currentOwner, null);
                }
            };
```

Content.Client/_Triad/ContrabandPermit/ContrabandPermitConsoleWindow.xaml.cs
```diff
@@ -0,0 +1,352 @@
...
+
+            label.SetMarkup(Loc.GetString("contraband-permit-no-permits"));
+
+            ActivePermitsTable.AddChild(label);
+
+            if (consoleComp.CurrentPermitReason != string.Empty && !PermitReasonTextBox.HasKeyboardFocus())
+                PermitReasonTextBox.Text = consoleComp.CurrentPermitReason;
+        }
+    }
```
rebaserHEAD, line 170: Draft-reason restore only happens when the permit list is empty; with any active permits on file the textbox loses its server-side draft on every state update. Move it out of the empty-list branch:

```suggestion
            ActivePermitsTable.AddChild(label);
        }

        if (consoleComp.CurrentPermitReason != string.Empty && !PermitReasonTextBox.HasKeyboardFocus())
            PermitReasonTextBox.Text = consoleComp.CurrentPermitReason;
    }
```

Content.Client/_Triad/ContrabandPermit/ContrabandPermitEntryContainer.xaml.cs
```diff
@@ -0,0 +1,158 @@
...
+
+            newButton.OnPressed += args =>
+            {
+                _window?.UpdateFocus(PermitOwner, item);
+            };
+
+            table.AddChild(newButton);
+        }
+    }
+}
```
rebaserHEAD, line 206: Same recycled-row problem as the owner rows, doubled: reused buttons never get their `Text` or `Pressed` updated (only set at creation), and the closure captures the creation-time `item`, so after a revoke shifts the list, a button shows one item and focuses another. Bind by index and sync reused buttons:

```suggestion
        // Make new buttons if needed
        if (index >= table.ChildCount)
        {
            var itemIndex = index;
            var newButton = new Button
            {
                ToggleMode = true,
                HorizontalExpand = true,
                Group = _permitItemButtons
            };

            newButton.OnPressed += args =>
            {
                if (itemIndex < PermitItems.Count)
                    _window?.UpdateFocus(PermitOwner, PermitItems[itemIndex]);
            };

            table.AddChild(newButton);
        }

        // Keep reused buttons in sync with the current item list
        if (table.GetChild(index) is Button button)
        {
            button.Text = itemMeta.EntityName;
            button.Pressed = isSelectedItem;
        }
    }
}
```

Resources/Textures/_Triad/Structure/Machines/computers.rsi/meta.json
```diff
@@ -0,0 +1,61 @@
+{
+  "version": 1,
+  "license": "CC-BY-SA-3.0",
+  "copyright": "Taken from tgstation at commit https://github.com/tgstation/tgstation/commit/bd6873fd4dd6a61d7e46f1d75cd4d90f64c40894.",
```
rebaserHEAD: Is the permits screen actually derived from tg at that commit, or was this copied from another meta.json as a template? If the sprite is original work it should carry your credit and the fork's license convention instead.

rebaserHEAD (changes requested):
Nice work on this thus far. We MIGHT need to revisit how we do the portraits as you're not wrong that it's certainly a way to go about it. As much as I like them I might suggest that we remove them just because they are sort of fragile. Otherwise the whole concept is very cool.

Content.Server/_Triad/ContrabandPermit/ContrabandPermitSystem.cs
TheRealMasterChief117: My thought process was behind this to make it urgent, so that players can't reset the warning by loading, but admin logs are permanent logs so no need!

Resources/Textures/_Triad/Structure/Machines/computers.rsi/meta.json
TheRealMasterChief117: made by me, remaking

Content.Server/_Triad/ContrabandPermit/ContrabandPermitSystem.cs
TheRealMasterChief117: Done (https://github.com/Triad-Sector/Triad_Sector/pull/454/changes/9a7377e325513d950520a5ea060c0d8c343ff43e)

Should fix it. Record keys are stored on the ``PermitRecordKey`` field on ``ContrabandPermitItemComponent ``. ``ContrabandPermitItemComponent`` is not removed until after the permit record is removed also.

Changed the ``AddPermitRecordToSectorService`` and ``RemovePermitRecordToSectorService`` methods to use Entity T for the permit items

Content.Server/_Triad/Humanoid/ClientHumanoidViewerSystem.cs
TheRealMasterChief117: Done, made it have a preview entity similar to the lobby UI.

Deletes itself when the control is disposed, and equips itself with a list of gear given by the server so dummies don't appear naked

Content.Server/_Triad/ContrabandPermit/ContrabandPermitSystem.cs
TheRealMasterChief117: Also done, server sends it through ``ContrabandPermitConsoleEntry``.``ContrabandPermitConsoleItem``

TheRealMasterChief117:
Let's go

PR #456, opened 2026-08-03
Rename MonoCoins to TriTokens, Swap Round-End Payment for Hourly

mnva0:
## About the PR
This PR renames MonoCoins to TriTokens, as well as making it so you are given 1 ~~MonoCoin~~ TriToken per hour. A notification is still given, which I'm not sure about

Also it adds a period to the balance notification (after the numbers) since I edited that line and it kind of annoyed me during testing

This was a difficult PR to make because I kept getting caught up in details. I'm editing the file quite a bit so I may as well make a new file, but I do not have the required experience to know how to do this well. The comments are outdated because I think it'd be odd updating the comments for local changes. But then again, the documentation for this specific file *probably won't* come up and I already left comments describing things anyhow

I'm also not sure about performance in any way because I don't currently know any method of measuring it. But this is theoretically just comparing numbers (in a way similar to the event scheduler) so I think it will be fine

## Why / Balance
1. We aren't Monolith
2. The round-end payment system is weird in that you don't even need to actually play to get 10 MonoCoins. You can just join 1 minute from round end and get the full 10 MonoCoins. This system might also kind of encourage playing longer than someone might intend. Unlike Monolith, we have ship-saving, so people should be encouraged to leave or join as they want

## Media
Taken as it rolled over. The "next time" variable was adjusted to five minutes here instead of an hour
image

## Requirements

## How to test
1. `vv /system/MonoCoinsSystem` to view (or edit) the next time for payout. When that time comes, it will be moved up
2. The `curtime` Toolshed command shows the time. `golobby` can be used to see the changes to the lobby menu, as well as test the fact it uses server time
3. I don't really know what to tell others to do from here but please break things.

**Changelog**
:cl: Minerva
- tweak: MonoCoins have been renamed to TriTokens.
- tweak: Instead of receiving 10 MonoCoins at round end, 1 TriToken is given hourly. This follows server time and not shift time.


rebaserHEAD:
I SWEAR I'M GONNA LOOK AT THIS THIS WEEKEND.

Content.Server/_Mono/MonoCoins/MonoCoinsSystem.cs
```diff
@@ -116,7 +140,7 @@ public async Task<bool> TrySubtractMonoCoinsAsync(NetUserId userId, int amount)
     /// <summary>
     /// Called when a round ends. Awards MonoCoins to players who appear in the station manifest.
     /// </summary>
-    private async void OnRoundEnd(RoundEndMessageEvent args)
+    private async void OnRoundEnd() // Triad - Removed argument. This is now called by Update.
```
rebaserHEAD: Rather than commenting that the names are inaccurate, rename them: `OnRoundEnd`, `AwardRoundEndMonoCoins` and `RoundEndReward` are all private so it's free. Something like `AwardHourlyPayout` / `HourlyReward`, with the summaries updated to match. A comment that exists should be true, local change or not.

Content.Server/_Mono/MonoCoins/MonoCoinsSystem.cs
```diff
@@ -25,8 +26,16 @@ public sealed class MonoCoinsSystem : EntitySystem
...
+    [Dependency] private readonly IGameTiming _gameTiming = default!; // Triad
 
-    private const int RoundEndReward = 10;
+    private const int RoundEndReward = 1; // Triad - 1<10. Name is inaccurate; pay is hourly instead of at round-end.
+
+    // Triad - Hourly instead of round-end.
+    private static readonly TimeSpan PayoutDelay = TimeSpan.FromSeconds(3600);
+
+    [DataField, ViewVariables(VVAccess.ReadWrite)]
+    public TimeSpan NextPayoutTime;
```
rebaserHEAD: `[DataField]` doesn't do anything on a system (they're not data definitions, nothing deserializes them) and it reads like this persists across restarts, which it doesn't. VV alone gives you the vv knob:
```suggestion
    [ViewVariables(VVAccess.ReadWrite)]
    private TimeSpan NextPayoutTime;
```

Content.Server/_Mono/MonoCoins/MonoCoinsSystem.cs
```diff
@@ -25,8 +26,16 @@ public sealed class MonoCoinsSystem : EntitySystem
     [Dependency] private readonly IChatManager _chatManager = default!;
     [Dependency] private readonly StationSystem _stationSystem = default!;
     [Dependency] private readonly StationRecordsSystem _stationRecords = default!;
+    [Dependency] private readonly IGameTiming _gameTiming = default!; // Triad
 
-    private const int RoundEndReward = 10;
+    private const int RoundEndReward = 1; // Triad - 1<10. Name is inaccurate; pay is hourly instead of at round-end.
+
+    // Triad - Hourly instead of round-end.
+    private static readonly TimeSpan PayoutDelay = TimeSpan.FromSeconds(3600);
```
rebaserHEAD: nit:
```suggestion
    private static readonly TimeSpan PayoutDelay = TimeSpan.FromHours(1);
```

Content.Server/_Mono/MonoCoins/MonoCoinsSystem.cs
```diff
@@ -130,7 +154,7 @@ private async void OnRoundEnd(RoundEndMessageEvent args)
             }
             else
             {
-                Logger.Debug($"Player {session.Name} ({session.UserId}) not found in station manifest, skipping MonoCoins reward");
+                Logger.Debug($"Player {session.Name} ({session.UserId}) not found in station manifest, skipping TriToken reward"); // Triad - MonoCoin rename.
```
rebaserHEAD: Since you're touching these lines anyway: `Logger.*` is obsolete, `Log.Debug/Info/Error` (the system's own sawmill) clears the deprecation warning for free. Same on the other two log lines you edited.

rebaserHEAD (commented):
Overall pretty good! Just a few nitpicks, honestly. Clean these up and I think we're good to go!

rebaserHEAD (changes requested):

mnva0:
There are some unrelated errors preventing me from testing this and thus making further commits. I have posted the errors in our maintainer channel

Content.Server/_Mono/MonoCoins/MonoCoinsSystem.cs
mnva0: Good catch! I didn't find that from a search when I was initially changing this

MantasDab360:
500 merge conflicts

Content.Server/_Mono/MonoCoins/MonoCoinsSystem.cs
```diff
@@ -40,8 +49,23 @@ public override void Initialize()
...
+    }
+
+    // Triad - Hourly instead of round-end.
+    public override void Update(float frameTime)
+    {
+        base.Update(frameTime);
+
+        var curTime = _gameTiming.CurTime;
+
+        if (curTime < NextPayoutTime)
```
rebaserHEAD: One design thing worth settling before this lands, and I'd rather raise it now than after it's live.

This is one global timer, so everyone in the manifest is paid on the same tick. Someone who joins two minutes before the tick gets a full TriToken, someone who plays 55 minutes and disconnects gets nothing. That's the same shape as the round-end flaw the description is arguing against, just at a tenth the size. `CurTime` also zeroes when the process restarts, and we restart at round end whenever a build goes out, so on those days everyone online loses whatever they'd accrued.

Ranked, most honest first:

1. Accrue per player off `PlayTimeTrackingManager.GetOverallPlaytime(session)` and pay when someone crosses each hour, persisting the last-paid mark next to the balance. Restart-proof, no join-timing exploit, and the tracker is already sitting there. Cost is one migration on both providers.
2. Keep the tick but make it per-session. `OnPlayerAttached` / `OnPlayerDetached` are empty stubs right now, so stash a connect time per user and pay each one on their own boundary. Kills the join-timing thing, still resets on restart, no schema change.
3. As written. Cheapest, and for a single token it's not the end of the world, but it keeps the behaviour the PR is objecting to.

I'd take 2 if you want this merged soon and 1 if you want it right. Say which and I'll stop poking.

Separate small thing on the same lines: `game.auto_pause_empty` defaults on, so `CurTime` freezes while the server is empty. That's almost certainly what you want, but it means the hour is an hour of occupied server rather than wall clock, and the changelog line about server time reads a bit stronger than that.

Content.Server/_Mono/MonoCoins/MonoCoinsSystem.cs
```diff
@@ -179,11 +203,11 @@ private async Task AwardRoundEndMonoCoins(ICommonSession session)
     {
         try
         {
-            var newBalance = await _db.AddMonoCoinsAsync(session.UserId, RoundEndReward);
-            Log.Info($"Awarded {RoundEndReward} MonoCoins to player {session.Name} ({session.UserId}). New balance: {newBalance}");
+            var newBalance = await _db.AddMonoCoinsAsync(session.UserId, HourlyReward); // Triad - Hourly instead of round-end.
```
rebaserHEAD: Still `AwardRoundEndMonoCoins`, and the summary above it still says "Awards round end MonoCoins". Same note as last round: it's private, so the rename is free, and a comment that exists has to be true. `AwardHourlyPayout` plus a summary along the lines of "Awards the hourly TriToken to one player" and it's done.

Content.Server/_Mono/MonoCoins/MonoCoinsSystem.cs
```diff
@@ -25,8 +26,16 @@ public sealed partial class MonoCoinsSystem : EntitySystem
...
+    [Dependency] private IGameTiming _gameTiming = default!; // Triad
 
-    private const int RoundEndReward = 10;
+    private const int HourlyReward = 1; // Triad - 1<10. Hourly instead of round-end.
+
+    // Triad - Hourly instead of round-end.
+    private static readonly TimeSpan PayoutDelay = TimeSpan.FromHours(1);
+
+    [ViewVariables(VVAccess.ReadWrite)]
+    public TimeSpan NextPayoutTime;
```
rebaserHEAD: VV reads private fields fine and nothing outside the system touches this, so `private` is enough.

```suggestion
    [ViewVariables(VVAccess.ReadWrite)]
    private TimeSpan NextPayoutTime;
```

Unrelated, since it's all through the diff: house marker style here is `// Triad:` with a colon (762 of those against 54 of the dash form).

Content.Server/_Mono/MonoCoins/MonoCoinsSystem.cs
```diff
@@ -179,11 +203,11 @@ private async Task AwardRoundEndMonoCoins(ICommonSession session)
...
         try
         {
-            var newBalance = await _db.AddMonoCoinsAsync(session.UserId, RoundEndReward);
-            Log.Info($"Awarded {RoundEndReward} MonoCoins to player {session.Name} ({session.UserId}). New balance: {newBalance}");
+            var newBalance = await _db.AddMonoCoinsAsync(session.UserId, HourlyReward); // Triad - Hourly instead of round-end.
+            Log.Info($"Awarded {HourlyReward} TriToken to player {session.Name} ({session.UserId}). New balance: {newBalance}"); // Triad - Hourly instead of round-end. TriToken rename.
 
             // Notify the player via chat
-            var notificationMessage = $"Round ended! You earned {RoundEndReward} MonoCoins. Your new balance: {newBalance}";
+            var notificationMessage = $"You earned {HourlyReward} hourly TriToken. Your new balance: {newBalance}."; // Triad - Hourly instead of round-end. TriToken rename.
```
rebaserHEAD: "You earned 1 hourly TriToken." is a bit of a mouthful. "You earned 1 TriToken. Your new balance: 12." says the same thing, and the hourly part explains itself by the second one.

Also the three log lines in this method each carry the same "Hourly instead of round-end. TriToken rename." trailer. One marker on the method is enough to make a future Mono merge conflict here, the repeats are just noise.

Content.Client/Lobby/LobbyUIController.cs
```diff
@@ -113,7 +113,7 @@ private void UpdateMonoCoinsDisplay()
             return;
 
         var balance = _monoCoins?.GetLastKnownBalance() ?? -1;
-        PreviewPanel.SetMonoCoinsText($"MonoCoins: {balance}");
+        PreviewPanel.SetMonoCoinsText($"TriTokens: {balance}"); // Triad - Renamed MonoCoin.
```
rebaserHEAD: The lobby label is the visible half, but the console commands are the other half and they all still say MonoCoins: `currency:balance` ("{} has {} MonoCoins"), `currency:add`, `currency:subtract` and `currency:transfer` have six of these strings between them in `Content.Server/_Mono/MonoCoins/`. Worth folding in here so we're not renaming the same currency twice.

PR #463, opened 2026-08-04
Radio Emotes + Possessive Emotes

nanbyt:
## About the PR

Cherypicked the ability to send emotes over radio, as well as emoting possessively without being spaced weird, from Delta-V
This also adds "?" as shorthand for "looks confused"

## Why / Balance

Must miao on radio,,,

## Media

Image from original.
image
Demonstration video from original.
https://github.com/user-attachments/assets/e070d97f-12c3-4042-8a36-f6bab7d00199

## Requirements

## How to test

Emote over radio using @![radio prefix] ex, @!; or @!:d
Emote possessively using @'    ("s" not needed, or else youll look a fool)

## Breaking changes

I have touched the forbidden chat code..

**Changelog**

:cl:
- add: ability to emote over radio using @![radio channel]
- add: ability to emote possessively using @'
- add: "?" shorthand emote for "looks confused"

### DIFF
--- a/Content.Server/Radio/EntitySystems/HeadsetSystem.cs
+++ b/Content.Server/Radio/EntitySystems/HeadsetSystem.cs
@@ -62,13 +62,6 @@ public sealed class HeadsetSystem : SharedHeadsetSystem
             && TryComp(headset.Headset, out EncryptionKeyHolderComponent? keys)
             && keys.Channels.Contains(channel.ID))
         {
-            // if (!TryComp<HandsComponent>(uid, out var hands) || hands.Count < 1 ||
-            //     TryComp<CuffableComponent>(uid, out var cuffable) && _cuffable.IsCuffed((uid, cuffable)))
-            // {
-            //     _popup.PopupEntity(Loc.GetString("headset-cant-reach"), uid, uid, PopupType.SmallCaution);
-            //     return false;
-            // }
-
             return true;
         }
 
--- a/Content.Server/Radio/EntitySystems/HeadsetSystem.cs
+++ b/Content.Server/Radio/EntitySystems/HeadsetSystem.cs
@@ -68,7 +68,7 @@ public sealed class HeadsetSystem : SharedHeadsetSystem
         return false;
     }
 
-    private void OnAudibleEmote(EntityUid uid, WearingHeadsetComponent component, EntityAudiblyEmotedEvent args)
+    private void OnAudibleEmote(EntityUid uid, WearingHeadsetComponent component, ref EntityAudiblyEmotedEvent args)
     {
         if (CheckRadioCapable(uid, component, args.Channel))
         {
--- a/Content.Server/Radio/EntitySystems/HeadsetSystem.cs
+++ b/Content.Server/Radio/EntitySystems/HeadsetSystem.cs
@@ -62,12 +62,12 @@ public sealed class HeadsetSystem : SharedHeadsetSystem
             && TryComp(headset.Headset, out EncryptionKeyHolderComponent? keys)
             && keys.Channels.Contains(channel.ID))
         {
-            if (!TryComp<HandsComponent>(uid, out var hands) || hands.Count < 1 ||
-                TryComp<CuffableComponent>(uid, out var cuffable) && _cuffable.IsCuffed((uid, cuffable)))
-            {
-                _popup.PopupEntity(Loc.GetString("headset-cant-reach"), uid, uid, PopupType.SmallCaution);
-                return false;
-            }
+            // if (!TryComp<HandsComponent>(uid, out var hands) || hands.Count < 1 ||
+            //     TryComp<CuffableComponent>(uid, out var cuffable) && _cuffable.IsCuffed((uid, cuffable)))
+            // {
+            //     _popup.PopupEntity(Loc.GetString("headset-cant-reach"), uid, uid, PopupType.SmallCaution);
+            //     return false;
+            // }
 
             return true;
         }
--- a/Content.Server/Radio/EntitySystems/HeadsetSystem.cs
+++ b/Content.Server/Radio/EntitySystems/HeadsetSystem.cs
@@ -10,6 +10,7 @@ using Content.Shared.Chat;
 using Content.Shared.Radio.EntitySystems;
 using Robust.Shared.Network;
 using Robust.Shared.Player;
+using Content.Shared._DV.Chat;
 
 namespace Content.Server.Radio.EntitySystems;
 
@@ -27,6 +28,7 @@ public sealed class HeadsetSystem : SharedHeadsetSystem
         SubscribeLocalEvent<HeadsetComponent, EncryptionChannelsChangedEvent>(OnKeysChanged);
 
         SubscribeLocalEvent<WearingHeadsetComponent, EntitySpokeEvent>(OnSpeak);
+        SubscribeLocalEvent<WearingHeadsetComponent, EntityAudiblyEmotedEvent>(OnAudibleEmote); // DeltaV
 
         SubscribeLocalEvent<HeadsetComponent, EmpPulseEvent>(OnEmpPulse);
     }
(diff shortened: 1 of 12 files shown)


Resources/Locale/en-US/_DV/headset/headset-component.ftl
```diff
@@ -0,0 +1,9 @@
+chat-radio-message-audible-emote-wrap = [color={$color}]{$channel} {$name} {$message}[/color]
+chat-radio-message-audible-possessive-emote-wrap = [color={$color}]{$channel} {$name}'s {$message}[/color]
+
+chat-radio-justice = Justice
+chat-radio-prison = Prison
+chat-radio-admeme-1 = Channel 1
+chat-radio-admeme-2 = Channel 2
+chat-radio-admeme-3 = Channel 3
+chat-radio-rootsong = Rootsong
```
mnva0: These channels are unused and should likely be removed since they aren't being cherrypicked here

Content.Server/Radio/EntitySystems/RadioSystem.cs
```diff
@@ -275,16 +290,30 @@ private string WrapRadioMessage(
...
-            ("color", channel.Color),
-            ("languageColor", languageColor),
-            ("fontType", language.SpeechOverride.FontId ?? speech.FontId),
-            ("fontSize", language.SpeechOverride.FontSize ?? speech.FontSize),
-            ("verb", Loc.GetString(_random.Pick(speech.SpeechVerbStrings))),
-            ("channel", $"\\[{channel.LocalizedName}\\]"),
-            ("name", name),
-            ("message", message),
-            ("language", languageDisplay));
+        // Triad Moved from SendRadioMessage
```
mnva0: If this was intended to be a block comment (above and below the changed content), it has no matching `End Triad` (or similar) at the bottom of the changes here

Minor nitpick: There should be some kind of separator between `Triad` and the actual comment. I personally use dashes `Triad - Changed fun.`

Content.Server/Radio/EntitySystems/RadioDeviceSystem.cs
```diff
@@ -207,6 +208,13 @@ private void OnReceiveRadio(EntityUid uid, RadioSpeakerComponent component, ref
 
         // log to chat so people can identity the speaker/source, but avoid clogging ghost chat if there are many radios
         var message = args.OriginalChatMsg.Message; // The chat system will handle the rest and re-obfuscate if needed.
+
+        // Triad please dont speak emotes thanks
```
mnva0: Minor nitpick: There should be some kind of separator between Triad and the actual comment. I personally use dashes `Triad - Changed fun.`

mnva0 (dismissed):
Sorry if a review took a while. I had a look through every file and found some little things. I would've done a more thorough review but I'll leave that to the other maintainers who know C# better than me

nanbyt:
Kaay! Removed extra lines and nittypicked the line commets

Content.Shared/Chat/SharedChatSystem.Emote.cs
```diff
@@ -208,6 +210,48 @@ public bool AllowedToUseEmote(EntityUid source, EmotePrototype emote)
...
+        }
+
+        var emoteType = input[0];
+        output = input[1..].TrimStart();
+
+        if (emoteType == AudibleEmotePrefix)
+        {
+            emoteType = input[1]; // Check for if we want AudiblePossessiveEmote
+            type = EmoteType.Audible;
+        }
```
rebaserHEAD, line 241: Crash on a bare `!`. Length is 1, so we get past the empty check and past the StartsWith check, then `input[1]` throws IndexOutOfRange. Typing `*!` in chat is enough to trigger it.

```suggestion
        if (emoteType == AudibleEmotePrefix)
        {
            type = EmoteType.Audible;
            if (input.Length > 1)
                emoteType = input[1]; // Check for if we want AudiblePossessiveEmote
        }
```

`source` is also unused in this method, worth dropping from the signature.

Content.Server/Radio/EntitySystems/HeadsetSystem.cs
```diff
@@ -51,13 +53,43 @@ private void UpdateRadioChannels(EntityUid uid, HeadsetComponent headset, Encryp
...
+            //     return false;
+            // }
+
+            return true;
+        }
+
+        return false;
+    }
+
+    private void OnAudibleEmote(EntityUid uid, WearingHeadsetComponent component, EntityAudiblyEmotedEvent args)
```
rebaserHEAD: `EntityAudiblyEmotedEvent` is a `[ByRefEvent]` record struct, but this subscribes by value, so this handler gets a copy and `args.Channel = null` never reaches the raiser. Nothing in the engine catches it: the ByRef analyzer only checks raise sites, and `EntAddSubscription` doesn't record ref-ness. It compiles and boots, it just silently doesn't de-dupe.

Symptom: anything with both a headset and an intrinsic transmitter on the same channel (IPCs, borgs) sends the audible emote over the radio twice. `OnSpeak` gets away with the same shape only because `EntitySpokeEvent` is a class.

```suggestion
    private void OnAudibleEmote(EntityUid uid, WearingHeadsetComponent component, ref EntityAudiblyEmotedEvent args)
```

Content.Server/Radio/EntitySystems/RadioSystem.cs
```diff
@@ -90,6 +92,16 @@ private void OnIntrinsicReceive(EntityUid uid, IntrinsicRadioReceiverComponent c
         }
     }
 
+    // DeltaV
+    private void OnIntrinsicAudibleEmote(EntityUid uid, IntrinsicRadioTransmitterComponent component, EntityAudiblyEmotedEvent args)
```
rebaserHEAD: Needs `ref` too, or it reads a stale copy once the headset handler starts nulling the channel correctly.

```suggestion
    private void OnIntrinsicAudibleEmote(EntityUid uid, IntrinsicRadioTransmitterComponent component, ref EntityAudiblyEmotedEvent args)
```

Content.Server/Radio/EntitySystems/RadioDeviceSystem.cs
```diff
@@ -207,6 +208,13 @@ private void OnReceiveRadio(EntityUid uid, RadioSpeakerComponent component, ref
 
         // log to chat so people can identity the speaker/source, but avoid clogging ghost chat if there are many radios
         var message = args.OriginalChatMsg.Message; // The chat system will handle the rest and re-obfuscate if needed.
+
+        // Triad - Prevent speaking emotes
+        if (args.emType == EmoteType.Audible || args.emType == EmoteType.AudiblePossessive)
+            {_chat.TrySendInGameICMessage(uid, message, InGameICChatType.Emote, ChatTransmitRange.GhostRangeLimitNoAdminCheck,
+            nameOverride: name, checkRadioPrefix: false, languageOverride: args.Language);
+            return;}
```
rebaserHEAD: Brace style, and this re-emits through the plain emote wrap so a possessive emote loses its `'s` when it comes out of a speaker. Fine to leave the fidelity bit, but the braces should go.

```suggestion
        // Triad - Prevent speaking emotes
        if (args.emType == EmoteType.Audible || args.emType == EmoteType.AudiblePossessive)
        {
            _chat.TrySendInGameICMessage(uid, message, InGameICChatType.Emote, ChatTransmitRange.GhostRangeLimitNoAdminCheck,
                nameOverride: name, checkRadioPrefix: false, languageOverride: args.Language);
            return;
        }
```

Content.Server/Radio/EntitySystems/HeadsetSystem.cs
```diff
@@ -51,13 +53,43 @@ private void UpdateRadioChannels(EntityUid uid, HeadsetComponent headset, Encryp
...
+            && keys.Channels.Contains(channel.ID))
+        {
+            // if (!TryComp<HandsComponent>(uid, out var hands) || hands.Count < 1 ||
+            //     TryComp<CuffableComponent>(uid, out var cuffable) && _cuffable.IsCuffed((uid, cuffable)))
+            // {
+            //     _popup.PopupEntity(Loc.GetString("headset-cant-reach"), uid, uid, PopupType.SmallCaution);
+            //     return false;
+            // }
+
+            return true;
```
rebaserHEAD: Dead block. If the reach check isn't wanted here, drop it rather than carrying commented DeltaV code we never ran. The "comment it out" rule is for upstream code we're diverging from, not for incoming ports.

```suggestion
            return true;
```

Content.Server/Radio/RadioEvent.cs
```diff
@@ -16,7 +17,8 @@ public readonly record struct RadioReceiveEvent(
     ChatMessage OriginalChatMsg,
     ChatMessage LanguageObfuscatedChatMsg,
     LanguagePrototype Language,
-    EntityUid RadioSource
+    EntityUid RadioSource,
+    EmoteType? emType //Triad
```
rebaserHEAD: Positional record members should be PascalCase, and this one is public API that other systems now read.

```suggestion
    EntityUid RadioSource,
    EmoteType? EmType //Triad
```

Call sites to follow: `args.emType` in `RadioDeviceSystem.OnReceiveRadio`.

rebaserHEAD (changes requested):
Needs some work and then a final review. Thanks for your efforts on this one! Chat is a real pain.

nanbyt:
Gulp, hopefully I didnt break anything. >< Thanks for catching the oopsies

mnva0 (commented):
GitHub is still saying I need to review this but I already did a very basic review (unless my review was requested again somewhere I couldn't see?). I am commenting this in an attempt to make GitHub realize the fact I have reviewed this

mnva0 (approved):
I'm not sure why my review was rerequested, but I had a quick look through again. Looks fine minor-nitpickwise to me from my extremely superficial review, but I'm unqualified to review C# PRs

rebaserHEAD:
First, sorry for the slow roll on this. I appreciate your patience while I was working on a few other things.

Secondly, I will say after considering this for a bit, that it's a bit odd to be able to emote over the radio which implies it's audio based in game. I might have a bit of a discussion with the other maintainers to chew this over. I'm on the fence with it leaning towards "No" but I can be convinced otherwise.

nanbyt:
I mean, they are referred to as radios in pretty much every instance. Characters wear headsets with visible microphones on them. I havent heard of a radio that _isnt_ talked into, or even audio based since morse ham radio is also auditory. (Im pretty sure sign language is even specifically not able to be 'spoken' on radio.) And while yes, this would allow people to emote physical actions into the radio, the idea for me is to just make it look nicer and without needing to make it appear like characters are saying "\*meows*" or something.

PR #464, opened 2026-08-04
Add TriTalk, a PDA Messaging App

rebaserHEAD:
## About the PR
Ports Delta-V's NanoChat and rebrands it TriTalk, a PDA messaging app. An ID card issued at spawn carries a 4 digit number, and you can message any other number from the PDA.

Delivery runs through the existing radio gates rather than its own path: the cartridge carries `ActiveRadio` on Common and goes through `RadioSendAttemptEvent`, `RadioReceiveAttemptEvent` and `HasActiveServer`, so a message reaches exactly as far as a headset would. Flying out on an expedition drops you off the sector network until you come back.

## Why / Balance
Quiet one-to-one comms for a sector where crews fly separate ships. Hopefully keeps the chatter out of broadband.

The directory only lists IDs issued to a player at spawn, and an unregistered card can't opt itself in, so spares, console printouts and forged cards stay off the books permanently. A stolen ID stays listed under its registered owner, which is what makes the directory worth anything as evidence. Messages log to admin logs under their own `LogType`, and a LogProbe pulls a card's history in round.

## Media
image

## Requirements
- [x] I have read relevant guidelines/documentation.

## How to test
Spawn two characters and examine each ID for its number, then message one from the other through Look up numbers. A spare ID pulled out of a locker should not appear in the lookup and its List number toggle should be greyed out. Unpower the telecomms server on the map and sends should start failing, with the lookup reporting no relay in range. LogProbe a card for its contacts and message history, and microwave one to scramble or wipe it. Builds clean and the server boots with no prototype errors, but none of this has been seen running yet, so it wants a proper in-game pass before merge.

## Breaking changes
New code sits under the `_DV` namespace with class names kept as `NanoChat*` so future Delta-V merges stay mechanical.

- `RadioSystem.HasActiveServer` is now public
- `LogProbeCartridgeSystem` is now partial
- `LogProbeUiFragment.UpdateState` takes the state object instead of `(string, List<PulledAccessLog>)`
- New `ActiveProgramChangedEvent` on `CartridgeLoaderSystem`
- New `NameIdentifierSystem.ReleaseUniqueName`

**Changelog**
:cl:
- add: TriTalk, a PDA messaging app. Examine an ID card to see its number. Needs a powered telecomms server in range, same as your headset.
- add: TriTalk lists issued IDs in a directory. You can unlist your own number.
- add: LogProbe can pull message history off a TriTalk card.


mnva0:
I'm a bit confused reading the description. The changelog and description imply that a range was added to headsets for all channels? Shortband is the only exception to "nigh-global radio" I've heard of

rebaserHEAD:
> I'm a bit confused reading the description. The changelog and description imply that a range was added to headsets for all channels? Shortband is the only exception to "nigh-global radio" I've heard of

It is TECHNICALLY correct since while you're on something like an expedition (which cuts you off from other chat channels) these also lose signal. That might be something I need to add actually... I don't think they will hold messages or indicate that they can't be sent when a player makes a map transition that doesn't have a comms server on it.

rebaserHEAD:
Moving this to a draft for now. Someone asked me if the app needs to be closed and re-opened to see the messages and after a bit of looking I'm finding that bug still exists. That said there is also an overhaul done to the PDA cartridge system that was done upstream that I think I'm also going to pull in from wizden which should also fix the notes app.

rebaserHEAD:
This is blocked by an engine upgrade to 285 or 286 so... Ugh.

PR #495, opened 2026-08-11
Stop Persisting Do-After Handles, Which Blocks Ship Saves

rebaserHEAD:
## About the PR

`DoAfterId` is a runtime handle with no type serializer, so writing one always throws. Eight components marked it `[DataField]` anyway, which hard-fails map serialization the moment any of them holds a non-null value. On a live server running this code it is currently blocking a player from saving their ship.

## Why / Balance

No balance impact. This is a data-loss fix.

A do-after id is a pointer into a user's live do-after list (`record struct DoAfterId(EntityUid Uid, ushort Index)`), with no `[DataDefinition]` and no serializer. It is meaningless across a map load, so it can only ever throw or write garbage. It should not have been persisted in the first place.

The failure looks like this:

```
[ERRO] entity_serializer: Caught exception while serializing component MagicMirror
       of entity barber scissors (10487741/n10487741, BarberScissors)
[ERRO] hardlight: Ship save failed for 'CRSV CENTRAL PROTOCOL 242-A0' on grid 10487225:
       System.ArgumentException: No type serializer or data definition found for type
       Content.Shared.DoAfter.DoAfterId when writing
```

`MagicMirror` also never cleared its handle when a do-after resolved. The only clears sat at the top of the *next* attempt, so any mirror or pair of barber scissors stayed permanently non-null after its first use. That is what arms the save-blocker: one used pair of scissors poisons every save of the grid it sits on, and it stays poisoned. The other seven are the same latent trap, they just need the right entity to land on a saveable grid.

The eight fields: `MagicMirror.DoAfter`, `Envelope.EnvelopeDoAfter`, `Grave.HandDiggingDoAfter`, `MechGrabber.DoAfter`, `MechFork.DoAfter` (`_NF`), `PlantAnalyzer.DoAfter` (`_NF`), `CanEscapeInventory.DoAfter`, `NPCSteering.DoAfterId`. Each keeps `[ViewVariables]`, so admin VV inspection is unchanged, and none of them had `AutoNetworkedField` on that field, so networking is untouched.

One note on the diff style: the removed attributes are commented out with `// Triad:` markers rather than deleted, which is a downstream fork convention for surfacing merge conflicts. If you would rather have clean deletions here, say so and I will rewrite the branch.

## Media

N/A, no ingame visual change.

## Requirements

## How to test

1. Spawn a ship grid with a pair of barber scissors or a mirror on it.
2. Use it on someone, and let the do-after run to completion.
3. Save the ship through the shipyard console. Before this change that fails with the `ArgumentException` above and the save is lost. With this branch it saves.

The guard test also covers the general case:

```
dotnet test Content.IntegrationTests/Content.IntegrationTests.csproj \
  --filter "FullyQualifiedName~DoAfterIdNotPersistedTest"
```

`DoAfterIdNotPersistedTest` walks every registered component and fails on any data field the serializer could reach a `DoAfterId` through: directly, through a nullable, through a collection or array element, or through a nested data definition. It checks `DataFieldBaseAttribute`, so `[IncludeDataField]` is covered too, and it walks the base chain by hand since `GetFields` will not return private fields declared on a base type.

It ships with a control. An empty offender list only means something if the walk can actually find a handle, so a second test plants three shapes it must catch and one it must not. It was also falsified against the real registry: with the `[DataField]` put back on `MagicMirrorComponent` the sweep fails and names `MagicMirror.DoAfter (MagicMirrorComponent)`.

Not yet verified in-game. The steps above are the real check and have not been run, so "the save now succeeds" is a prediction backed by the compile, the reflection sweep, and the falsification, not a result.

## Breaking changes

None expected.

No prototype in `Resources/` sets any of the removed fields, so nothing starts erroring on load. Existing saved ships cannot contain these nodes either: writing one always threw, so the field could never have made it into a save file in the first place.

Worth flagging for review: the serialization generator emitted 50 read/write/validate sites for `DoAfterId` before this change and emits zero after. The only remaining use is `List<DoAfterId>` on `ShipRepairToolComponent`, which is `[ViewVariables]` only and was never serialized.

**Changelog**

:cl:
- fix: Ships carrying a used mirror or pair of barber scissors can be saved again.

### DIFF
--- a/Content.IntegrationTests/Tests/_Triad/Serialization/DoAfterIdNotPersistedTest.cs
+++ b/Content.IntegrationTests/Tests/_Triad/Serialization/DoAfterIdNotPersistedTest.cs
@@ -1,0 +1,220 @@
+// SPDX-FileCopyrightText: 2026 Triad Sector contributors
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using Content.Shared.DoAfter;
+using Robust.Shared.GameObjects;
+using Robust.Shared.Serialization.Manager.Attributes;
+using Robust.Shared.ViewVariables;
+
+namespace Content.IntegrationTests.Tests._Triad.Serialization;
+
+/// <summary>
+/// Guards the whole family of bugs behind a live ship-save outage: a <see cref="DoAfterId"/> reaching
+/// a persisted data field.
+///
+/// A do-after id is a runtime handle, a user plus an index into that user's live do-after list. It has
+/// no type serializer and no data definition, so the serializer throws "No type serializer or data
+/// definition found" the moment it is asked to write one. A component that marks such a field
+/// [DataField] therefore hard-fails map serialization as soon as the field goes non-null, taking the
+/// entire ship-grid save with it. In production this was MagicMirror on a pair of barber scissors,
+/// which blocked every save of the grid it sat on.
+///
+/// Eight components carried that mistake, all inherited. The point of this test is that the ninth
+/// cannot land quietly, including by arriving through an upstream merge.
+/// </summary>
+[TestFixture]
+[TestOf(typeof(DoAfterId))]
+public sealed class DoAfterIdNotPersistedTest
+{
+    [Test]
+    public async Task NoRegisteredComponentPersistsADoAfterHandle()
+    {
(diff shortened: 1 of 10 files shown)


TheRealMasterChief117 (approved):
Awesome

A lot of other things need this as well, fields that shouldn't be serialized at all. One of these poisoned fields could be a similar reason as to why flatpackers brick loading a ship

PR #526, opened 2026-08-19
Breaching Hammer Tweak

gaywolfgirl6:
## About the PR
Removed the 10% slow on the Breaching Hammer.

## Why / Balance
This is a fun melee weapon that feels bad to use from the slow provided, and doesn't provide a clear balancing reason for it having it to begin with.

## Requirements

## How to test
Find breaching hammer, enjoy a little bit of newfound freedom.

## Breaking changes
Removed the 6 lines of code that added the slow to the Breaching Hammer.

**Changelog**

:cl:
- tweak: Removed the movement speed debuff given upon wielding a Breaching Hammer

### DIFF
--- a/Resources/Prototypes/Nyanotrasen/Entities/Objects/Weapons/Melee/breaching_hammer.yml
+++ b/Resources/Prototypes/Nyanotrasen/Entities/Objects/Weapons/Melee/breaching_hammer.yml
@@ -23,12 +23,6 @@
     bluntStaminaDamageFactor: 1.30 # Triad decrease to 1.3
     soundHit:
       collection: MetalThud
-  - type: SpeedModifiedOnWield # Mono
-    walkModifier: 0.9
-    sprintModifier: 0.7
-  - type: ClothingSpeedModifier
-    walkModifier: 1
-    sprintModifier: 0.9
   - type: Wieldable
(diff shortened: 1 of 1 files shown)


Wuffindy:
this weapon deals one of the highest blunt & stamina damage in melee along with structural damage in case if it wasn't clear enough

Also you removed the unwielded speed alongside its wielded speed

gaywolfgirl6:
> this weapon deals one of the highest blunt & stamina damage in melee along with structural damage in case if it wasn't clear enough
> 
> Also you removed the unwielded speed alongside its wielded speed

It does! Currently, the Tintreach has more raw damage output as it has an added shock, and no movement penalty.

My only other real comparison for what this does, would be something else equally available to it. It's most commonly seen as an expedition treasure, so I compared it to another, the Bulldog. The bulldog has more range, faster TTK, can destroy plastit walls faster, while not providing such downsides. 

I've also not really seen anyone using a melee other than an energy sword. Even Oni, the race geared to it, seem to use 4 gauge shotguns over it.

If it becomes a problem, I have no issue reverting the change, but the added negatives ontop of being able to stamina crit yourself, something guns do not do, felt very superficial, thus why I made this PR.

PR #540, opened 2026-08-22
Fix/Ship Save Serialization

rebaserHEAD:
## About the PR

Ship saves have been failing on live for weeks: 24 distinct grids over the last 30 days, because a single component holding runtime state the YAML writer cannot express takes the whole save down with it. This fixes every cause found in that window plus two latent ones, and re-enables flatpackers for ship saving, which turned out to have been excluded with no technical reason on record.

## Why / Balance

The save runs with the engine default `EntityExceptionBehaviour.Rethrow`, so one bad component aborts the entire grid save. Seven distinct causes, all the same shape:

| Component | Why it threw |
|---|---|
| `SmartFridge.containedEntries` | dictionary keyed by a data definition, which cannot be a YAML mapping key |
| `KillTracker.lifetimeDamage` | same, `KillSource` is a polymorphic record |
| `NPCRetaliation.attackMemories` | keyed by `EntityUid` on arbitrary attackers; two off-grid keys both write `invalid` and collide |
| `PayloadTrigger.grantedComponents` | `HashSet<Type>`; `RuntimeType` has no data definition |
| `BodyPartAppearance` | a DataField whose YAML key was literally `type`, colliding with the component node's own discriminator |
| `DeviceLinkSource.linkedPorts` | the top cause on live: existence was the wrong test, see below |
| `OreSilo` | both halves of a two-way link were persisted |

`DeviceLinkSource` is worth calling out because a guard for it already existed and did not work. The save uses `MissingEntityBehaviour.Ignore`, and `EntitySerializer.Write` returns the literal string `invalid` for any `EntityUid` outside the set being serialized, **alive or not**. `LinkedPorts` is keyed by `EntityUid`, so a docking button wired to station-side shutters writes `invalid` for a perfectly live entity, and the second such link is a duplicate dictionary key. The old cleanup only looked for deleted entities, so it never saw this class at all.

The ore silo had accumulated six defensive `Exists()` checks at its read sites, one of them recording millions of errors a day. Only the client half is persisted now and the silo's client set is rebuilt from it on `ComponentStartup`, which is the shape the engine already uses for `DeviceLinkSinkComponent.LinkedSources`.

On the balance side, the only player-facing change is flatpackers. They were dropped from saves by `SavingContraband`, which they picked up in a batch captioned "obvious non-ship entities" under the heading "Uplinks and bundled items", next to mercenary uplinks and criminal-records computers, inside a commit about an unrelated bug. Their assigned class is the mildest one, "Legal to own, but will be seized upon storing". A flatpacker holding a board, carrying materials, mid-pack and linked to an off-grid ore silo saves and loads back intact, so the exclusion is lifted. Its machine board carried a separate marker and is lifted with it.

## Media

N/A, this is a save-pipeline fix with no visual component.

## Requirements

## How to test

Automated, 9 integration tests across three fixtures:

```
dotnet test Content.IntegrationTests/Content.IntegrationTests.csproj --filter "FullyQualifiedName~ShipSaveSerializationTest|FullyQualifiedName~FlatpackerSaveTest|FullyQualifiedName~OreSiloSaveTest"
```

Two of those are deliberately controls rather than assertions about our own code. `ControlTwoOffGridKeysInAPersistedDictionaryStillCollide` plants a component with the bug's exact shape and proves the serializer still throws on two off-grid keys and not on one, so a clean run of the rest means something. `ListingRebuildDependsOnTheSanitizerStrippingMapInit` saves the same fridge through the engine's own path, where the listing comes back empty over a full container, pinning why the Triad round trip works at all.

In game: buy a ship, save it, load it. Verified over three save/load cycles with no `Ship save failed`, no serializer exception, and no occurrence of the literal string `invalid` in the saved file.

Not yet verified in game, and worth a reviewer's eye: a ship carrying an actual flatpacker, and a silo-and-client pair on the same grid. Both are covered by integration tests but neither has been exercised by hand. For the silo specifically, put it on the ship rather than the station, since `CanTransmitMaterials` refuses any pair whose grids differ.

## Breaking changes

- `OreSiloComponent.Clients` is no longer a `[DataField]`. It stays `[AutoNetworkedField]` and is rebuilt on `ComponentStartup` from `OreSiloClientComponent.Silo`, which is now the authoritative half. Existing maps that set it in YAML would silently stop having it applied, though nothing in the repo does.
- `BodyPartAppearance`'s DataField key changed from `type` to `layerType`. Nothing sets it in YAML, and it could never round-trip before because writing it threw, so no data migration is needed.
- `SmartFridge.containedEntries`, `KillTracker.lifetimeDamage`, `NPCRetaliation.attackMemories` and `PayloadTrigger.grantedComponents` are no longer persisted. All four are runtime state and no prototype sets any of them. Stale keys in already-saved ships are ignored on read: the generated `Read` is a `TryGet` per known key with no pass over leftovers.
- New internals on `ShipyardGridSaveSystem`: `TryBuildShipSaveYaml` and `ShipSaveSerializationOptions`, both `internal`, split out so the fallible part of a save is reachable without a session or a signing key. `TrySaveGridAsShip` is behaviourally unchanged.
- New public API `SharedOreSiloSystem.ClearSiloLink`, so callers outside that system can drop a link without writing to `[Access]`-restricted members.

**Changelog**

:cl:
- fix: Ship saves no longer fail outright when a smart fridge, a wired docking button, an assembled grenade, a severed limb or an ore silo link is aboard.
- fix: Loading a saved ship no longer leaves a stocked smart fridge reporting itself empty.
- tweak: Flatpackers and their machine boards can now be saved with a ship instead of being seized.

### DIFF
--- a/Content.Server/_Triad/Shipyard/ShipyardGridSaveSystem.cs
+++ b/Content.Server/_Triad/Shipyard/ShipyardGridSaveSystem.cs
@@ -473,7 +473,7 @@ public sealed partial class ShipyardGridSaveSystem : EntitySystem
     /// whose sink is not on this grid. Preserves links whose sink is on the grid being saved.
     /// </summary>
     /// <remarks>
-    /// Triad: entity existence is NOT the right test here, which is why this ran for months while
+    /// Entity existence is NOT the right test here, which is why this ran for months while
     /// DeviceLinkSource stayed the top cause of failed ship saves. The save runs with
     /// <see cref="MissingEntityBehaviour.Ignore"/>, and EntitySerializer.Write returns the literal
     /// string "invalid" for ANY EntityUid outside the set being serialized, alive or not. Because
--- a/Content.Server/_Triad/Shipyard/ShipyardGridSaveSystem.cs
+++ b/Content.Server/_Triad/Shipyard/ShipyardGridSaveSystem.cs
@@ -276,66 +276,11 @@ public sealed partial class ShipyardGridSaveSystem : EntitySystem
 
         try
         {
-            // Purge invalid entities
-            PurgeInvalidEntities(gridUid);
-
-            // Triad: runs AFTER the purge so it sees the final entity set. In practice the engine
-            // already drops links to purged sinks (SharedDeviceLinkSystem.OnSinkRemoved fires on
-            // ComponentRemove during Del), so this ordering is defensive rather than load-bearing;
-            // the fix that matters is the off-grid test inside CleanupBrokenDeviceLinks.
-            // Clean up broken device links before serialization
-            CleanupBrokenDeviceLinks(gridUid);
+            // Triad: the part of a save that can actually fail lives in TryBuildShipSaveYaml, so it
+            // can be exercised without a session, a signing key or a client to deliver to.
+            if (!TryBuildShipSaveYaml(gridUid, out var yaml, out var appraisalCost))
+                return false;
             // End Triad
-
-            // remove any edge spreaders, we cannot save these
-            RemoveEdgeSpreaderComponentComponentsOnGrid(gridUid);
-
-            // Reset fabricators: disable loop/skip and cancel active jobs to prevent mid-save exceptions
-            ResetFabricatorsOnGrid(gridUid);
-
-            // reset station records computers to prevent errors with StationRecordsFilter
-            ResetGeneralRecordsConsolesOnGrid(gridUid);
-
-            // Remove repair data, it is re-added on load
-            RemComp<ShipRepairDataComponent>(gridUid);
-
-            // Remove SpreaderGrid component from grid;
-            RemComp<SpreaderGridComponent>(gridUid);
-
-            //_sawmill.Info($"Serializing ship grid {gridUid} as '{shipName}' after transient purge using direct serialization");
-
-            // 1) Serialize the grid and its children to a MappingDataNode (engine-standard format)
-            var entities = new HashSet<EntityUid> { gridUid };
-            // Prefer AutoInclude to pull in dependent entities; we'll sanitize nullspace and parents out below
-            var opts = SerializationOptions.Default with
-            {
-                // Do NOT auto-include referenced entities (players/admin observers/etc.).
-                // This prevents exceptions when encountering unserializable entities and keeps saves scoped to the grid.
-                MissingEntityBehaviour = MissingEntityBehaviour.Ignore,
-                ErrorOnOrphan = false,
-                // Disable auto-include logging to avoid excessive log spam/lag during saves.
-                LogAutoInclude = null
-            };
-
-
-            // triad start
-            // these three lines were lifted from the loading code, and should be refactored into a function at some point
-            var loadShipPrice = _configManager.GetCVar(TriadCCVars.LoadShipPrice);
-            var fullAppraisal = _pricing.AppraiseGrid(gridUid, null);
-            var appraisalCost = (int)MathF.Round((float)fullAppraisal * loadShipPrice);
-            // triad end
-
-            var (node, category) = _mapLoader.SerializeEntitiesRecursive(entities, opts);
-            /* if (category != FileCategory.Grid)
-            {
-                _sawmill.Warning($"Expected FileCategory.Grid but got {category}; continuing with sanitation");
-            } */
-
-            // 2) Sanitize the node to match blueprint conventions
-            SanitizeShipSaveNode(node);
-
-            // 3) Convert MappingDataNode to YAML text without touching disk
-            var yaml = WriteYamlToString(node);
 
             // Triad ship anti-tamper start
             var shipFileBox = _tamperPolicy.SignSave(yaml, appraisalCost);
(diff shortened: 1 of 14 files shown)


Content.Server/_Triad/Shipyard/ShipyardGridSaveSystem.cs
```diff
@@ -375,6 +327,92 @@ public bool TrySaveGridAsShip(EntityUid gridUid, string shipName, string playerU
...
+        // Triad: runs AFTER the purge so it sees the final entity set. In practice the engine
+        // already drops links to purged sinks (SharedDeviceLinkSystem.OnSinkRemoved fires on
+        // ComponentRemove during Del), so this ordering is defensive rather than load-bearing;
+        // the fix that matters is the off-grid test inside CleanupBrokenDeviceLinks.
+        // Clean up broken device links before serialization
+        CleanupBrokenDeviceLinks(gridUid);
+
+        // Same treatment for ore-silo links, which are the other two-way EntityUid link on a ship.
+        CleanupOffGridOreSiloLinks(gridUid);
+        // End Triad
```
TheRealMasterChief117: You do not need 'triad start' and 'end triad' on triad-specific files

Content.Server/_Triad/Shipyard/ShipyardGridSaveSystem.cs
```diff
@@ -375,6 +327,92 @@ public bool TrySaveGridAsShip(EntityUid gridUid, string shipName, string playerU
...
+
+        // 1) Serialize the grid and its children to a MappingDataNode (engine-standard format)
+        var entities = new HashSet<EntityUid> { gridUid };
+
+        // triad start
+        // these three lines were lifted from the loading code, and should be refactored into a function at some point
+        var loadShipPrice = _configManager.GetCVar(TriadCCVars.LoadShipPrice);
+        var fullAppraisal = _pricing.AppraiseGrid(gridUid, null);
+        appraisalCost = (int)MathF.Round((float)fullAppraisal * loadShipPrice);
+        // triad end
```
TheRealMasterChief117: Same here

Content.Server/_Triad/Shipyard/ShipyardGridSaveSystem.cs
```diff
@@ -431,52 +469,107 @@ private void ResetGeneralRecordsConsolesOnGrid(EntityUid gridUid)
     }
 
     /// <summary>
-    /// Cleans up broken device links where one or both linked entities no longer exist.
-    /// Preserves valid links where both source and sink entities are still present.
+    /// Cleans up device links that cannot be serialized: links whose sink no longer exists, and links
+    /// whose sink is not on this grid. Preserves links whose sink is on the grid being saved.
     /// </summary>
+    /// <remarks>
+    /// Triad: entity existence is NOT the right test here, which is why this ran for months while
```
TheRealMasterChief117:
```suggestion
    /// Entity existence is NOT the right test here, which is why this ran for months while
```

Content.Server/_Triad/Shipyard/ShipyardGridSaveSystem.cs
```diff
@@ -431,52 +469,107 @@ private void ResetGeneralRecordsConsolesOnGrid(EntityUid gridUid)
...
-            /* if (linksRemoved > 0)
-                _sawmill.Info($"CleanupBrokenDeviceLinks: Removed {linksRemoved} broken device link(s) from {sourcesProcessed} source(s) on grid {gridUid}"); */
         }
         catch (Exception e)
         {
             _sawmill.Warning($"CleanupBrokenDeviceLinks: Exception while cleaning device links on grid {gridUid}: {e.Message}");
         }
     }
 
+    // Triad start
```
TheRealMasterChief117:
```suggestion
```

Content.Server/_Triad/Shipyard/ShipyardGridSaveSystem.cs
```diff
@@ -431,52 +469,107 @@ private void ResetGeneralRecordsConsolesOnGrid(EntityUid gridUid)
...
+
+                _oreSilo.ClearSiloLink((uid, client));
+            }
+        }
+        catch (Exception e)
+        {
+            _sawmill.Warning($"CleanupOffGridOreSiloLinks: Exception while clearing silo links on grid {gridUid}: {e.Message}");
+        }
+    }
+    // Triad end
```
TheRealMasterChief117:
```suggestion
```

Resources/Prototypes/Entities/Objects/Devices/Circuitboards/Machine/production.yml
```diff
@@ -1142,8 +1142,11 @@
       MatterBin: 1 # Frontier stackRequirements<requirements
     stackRequirements:
       Steel: 1
-  - type: SavingContraband
-    examineText: ship-saving-contraband-1-text
+# Triad: re-enabled alongside MachineFlatpacker. The board carried its own marker, and because the
+# contraband check in IsInvalidEntity runs before the IsInsidePersistentStorage check, un-marking only
+# the machine left a flatpacker that saved and loaded with an empty board slot.
+#  - type: SavingContraband
+#    examineText: ship-saving-contraband-1-text
```
TheRealMasterChief117:
```suggestion
```
Just delete the entire comment, this is a wizden file

Resources/Prototypes/Entities/Structures/Machines/flatpacker.yml
```diff
@@ -97,8 +97,14 @@
...
-  - type: SavingContraband
-    examineText: ship-saving-contraband-1-text
+# Triad: re-enabled for ship saving. This was excluded with no technical justification on record: it
+# entered the exclusion list in a batch captioned "obvious non-ship entities", under the heading
+# "Uplinks and bundled items", alongside mercenary uplinks and criminal-records computers. A
+# flatpacker holding a board, carrying stored materials, mid-pack and linked to an off-grid ore silo
+# saves and loads back intact; see FlatpackerSaveTest. Its machine board carried a separate marker
+# and is re-enabled with it. Restore both lines together if this is ever reverted.
+#  - type: SavingContraband
+#    examineText: ship-saving-contraband-1-text
```
TheRealMasterChief117:
```suggestion
```

rebaserHEAD:
Fair. Gimme a sec to look it over real quick.

PR #560, opened 2026-08-27
Better Thruster Burn and Thruster Restriction Enforcement

TheRealMasterChief117:
## About the PR
- Thrusters now burn everything in the exhaust path, including structures (6 tiles)
- Popups for getting burnt by structure exhaust
- Past the 6 tile radius, a raycast checks if the path is clear in all cardinal directions. If atleast one path is clear, the thruster will work

Thruster exhaust damage now scales with distance, mobs take 35% less damage due to the range increase

## Why / Balance
So thruster restrictions can be enforced via gameplay rather than a rule.
The exhaust needs somewhere to go, we want realistic looking thrusters.

Also adds a neat mechanic for thruster armor.

In the 6-tile burn radius, you CAN heavily armor your thrusters.. however, the exhaust will slowly burn away the armor. So make your choice.

Grilles are NOT affected by thruster exhaust

## How to test
Test a thruster enclosed in a box. It should be able to thrust as long as the direction the nozzle is pointing isn't blocked by something past the burn radius (6 tiles)

If you put a wall 6 tiles away from the thruster nozzle and enclose it, it will not be able to thrust. 
If you remove one of the walls, it should be able to thrust. At least ONE SIDE needs to be fully exposed to space

## Media

https://github.com/user-attachments/assets/0f5f1723-022b-42f7-9d52-68e9f5a9e423

## Requirements

## Technical details
Major changes to the thruster system. ``NozzleExposed`` now passes a ``Entity<TransformComponent, ThrusterComponent>`` rather than just a ``TransformComponent``.

In the burn damage update, ``EntityLookUpSystem.GetLocalEntitiesIntersecting`` is done to check for entities intersecting with the hitbox (Anchored entities DO NOT trigger the start collide event, so this is needed so structures are damaged)

``ThrusterComponent.Colliding`` field now has ``ViewVariablesAttribute``, and is now a HashSet so duplicate entries are not done. In some cases, an entity would get added to Colliding multiple times, which would make them take as much damage as they were in the list. This fixes that issue. 

Four raycasts are sent out to check each cardinal direction (North, South, West, East)
Thrusters need at least one clear path to operate. If all 4 paths are closed, the thrusters won't turn on
Objects in front of a thruster's nozzle that are in burn distance don't count towards the raycast check because the thruster will burn it anyway

**Changelog**
:cl:
- add: Added popup and damage sounds for being burnt by thruster exhaust!
- tweak: Thruster exhaust now burns objects in a 6-tile distance, including structures. The damage scales with how close the object is to the exhaust. The exhaust damage for mobs still has the same 2-tile range
- add: Grilles and railings are immune to thruster exhaust damage.
- tweak: Thrusters now need a path to clear space to function. This path doesn't need to be the same direction as the nozzle, as long as one part of the thruster is exposed to space. Only windows, walls, and doors count towards blocking it. Grilles and other small structures DO NOT block this path.
- fix: Fixed thrusters exhaust damage sometimes dealing twice or three times as much damage than normal.


Eillyrea:
The other main thing with thruster rules, which I'm unsure if you can do is "Has to be shootable from space" to stop people building large walls that are just far enough away to miss exhausts and then encloses the thrusters or hides them from being hit.  

Idk if that's possible to include as an auto-shut off if it doesn't detect a clear route out to space that like, 70 degrees or more of open?

TheRealMasterChief117:
> The other main thing with thruster rules, which I'm unsure if you can do is "Has to be shootable from space" to stop people building large walls that are just far enough away to miss exhausts and then encloses the thrusters or hides them from being hit.
> 
> Idk if that's possible to include as an auto-shut off if it doesn't detect a clear route out to space that like, 70 degrees or more of open?

Working on it so the thrusters just need at least one part exposed to space

TheRealMasterChief117:
More changes:

raycast check Angles are now configurable in YML
You can offset the point 
image

Should be a lot more fair now since thrusters exposed to space diagonally will work as well

Resources/Prototypes/_Mono/Entities/Structures/Shuttles/capital_thrusters.yml
```diff
@@ -38,10 +38,29 @@
...
+    - angle: South
+      worldOffset: 0.5,-0.5
+    - angle: SouthWest
+      worldOffset: 0.5,-0.5
+    - angle: SouthEast
+      worldOffset: 0.5,-0.5
+    - angle: West
+      worldOffset: 0.5,-0.5
+    - angle: East
+      worldOffset: 0.5,-0.5
```
TheRealMasterChief117:
```suggestion
      originOffset: 0.5,-0.5
    - angle: NorthEast
      originOffset: 0.5,-0.5
    - angle: NorthWest
      originOffset: 0.5,-0.5
    - angle: South
      originOffset: 0.5,-0.5
    - angle: SouthWest
      originOffset: 0.5,-0.5
    - angle: SouthEast
      originOffset: 0.5,-0.5
    - angle: West
      originOffset: 0.5,-0.5
    - angle: East
      originOffset: 0.5,-0.5
```

Content.Server/Shuttles/Systems/ThrusterSystem.cs
```diff
@@ -39,11 +53,28 @@ public sealed partial class ThrusterSystem : EntitySystem
...
 
+    // Triad Start
+    private const CollisionGroup StructureMask = CollisionGroup.FullTileMask;
+    private const CollisionGroup BurnMask = CollisionGroup.FullTileMask;
+
+    private readonly HashSet<EntityUid> _toRemoveColliding = new();
+    private readonly HashSet<Entity<TransformComponent>> _fixtureLookupEnts = new();
+
+    private readonly Direction[] _thrusterBlockCheckDirections = [Direction.North, Direction.East, Direction.South, Direction.West];
+
```
TheRealMasterChief117:
```suggestion
```

Content.Server/Shuttles/Systems/ThrusterSystem.cs
```diff
@@ -22,6 +21,15 @@
...
 using Content.Server.Construction; // Frontier
 using Content.Shared.DeviceLinking.Events; // Frontier
+using Robust.Shared.Physics; // Triad
+using Content.Shared.Whitelist; // Triad
+using Content.Shared.Mobs.Components; // Triad
+using Robust.Shared.Audio.Systems; // Triad
+using Content.Shared.Popups; // Triad
+using Content.Shared.IdentityManagement; // Triad
+using Robust.Shared.Player;
+using System.Linq;
```
TheRealMasterChief117:
```suggestion
using Robust.Shared.Player; // Triad
using System.Linq; // Triad
```

PR #565, opened 2026-08-28
Cherrypick: Cherrypick: Frontier's Port of DV's Tape Recorders (#4170)

mnva0:
## About the PR
This PR cherrypicks https://github.com/Monolith-Station/Monolith/pull/4170, which is a cherrypick of NF's cherrypick of tape recorders

Original PR description:
```
This PR cherrypicks the following PRs from Frontier. I also checked out the most recent NF version of some files to get fixes
- https://github.com/new-frontiers-14/frontier-station-14/pull/2623
- https://github.com/new-frontiers-14/frontier-station-14/pull/4177

I've opted to change the loadout price from 600 to 0 (consistency with binoculars in the same category) and change the loadout item from TapeRecorder to TapeRecorderFilled (starting useless is kind of odd for loadout options? especially compared to binoculars in the same category)
I also moved it to `tools.yml` (and moved the tape recorder to Tools) under a Mono namespace because the file it was in didn't exist
```

There is a minor visual problem where tape recorders that are saved and loaded don't *look* like they have a tape in them. This is fixed by just taking the tape out (it's still there!) and reinserting it

Mode persists, but I'm not certain about how to fix it if it is a problem. Maybe I can make the tape recorder reset to Pause on initialization?

## Why / Balance
Original PR description:
```
This *does* unfortunately make translating languages easier (ListenEvent does not support languages), but I think it'd add far more in terms of fun RP shenanigans than it'd take away.
Holopads already translate languages (same case. ListenEvent) anyway. If it's required before this gets merged, I could try to add language support to ListenEvent
```

Tape recorders add a lot of gimmick space into the game. They're cool!

## Media
image

image

image

Seems to work fine in persistence.
image

image

## Requirements

- [X] I have read the guidelines and documentation relevant to this PR.

## How to test
Original testing instructions:
```
1. Spawn in a tape recorder. Put in a tape if you didn't spawn the filled version
2. Play around with it. Try other languages (add using `self language:add Draconic true true` or something) if you'd like
3. Spawn in four buttons (or signalers?) to test signal control. Link all 4 buttons to the ports on the tape recorder
4. Pick up the tape recorder and open the UI. Press the new signal buttons. The tape recorder updates its UI and changes mode
5. Remove the tape. Try the buttons again. Nothing
6. Assuming you recorded something on the tape, damage it by hitting it and repair it with a screwdriver. Play the tape. It corrupted the audio

.

1. Open the loadout menu in the Tools category
2. Spawn with a tape recorder
3. It starts filled
```

1. Mess around with persistence and make sure nothing goes wrong. Transcripts, tapes without a tape, tape recorders
2. Try damaging the tapes and seeing if they persist
3. Just try to break things
4. Check to make sure tape colors also persist. I didn't test this and I'm already onto the next project and I heard something like this happened with something else recently

**Changelog**
:cl: Minerva
- add: Tape recorders have been ported from Frontier for recording speech. They can be printed in Autolathes or selected in loadout under Tools. They can also be signal-controlled by using four signal ports corresponding to their buttons.


rebaserHEAD:
Will review this when I get home tonight but this is FUN.

Content.Shared/_DV/TapeRecorder/Components/TapeRecorderComponent.cs
```diff
@@ -0,0 +1,101 @@
...
+    public SoundSpecifier RewindSound = new SoundPathSpecifier("/Audio/_DV/Items/TapeRecorder/rewind.ogg")
+    {
+        Params = AudioParams.Default.WithVolume(-2f).WithMaxDistance(3f)
+    };
+
+    /// <summary>
+    /// Ports for signal control
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public ProtoId<SinkPortPrototype> PausePort = "Pause";
```
rebaserHEAD: The entity's DeviceLinkSink declares `TapePause` (the rename was right, plain `Pause` collides with the lathe pause port), but this default still says `Pause` and nothing in the yaml overrides it. `OnSignalReceived` compares against this field, so the pause link can be wired up in the linking UI but the signal never matches: pause via signal is dead wire while the other three ports work.

```suggestion
    public ProtoId<SinkPortPrototype> PausePort = "TapePause";
```

Resources/Prototypes/Roles/Jobs/Wildcards/reporter.yml
```diff
@@ -4,12 +4,12 @@
   description: job-description-reporter
   playTimeTracker: JobReporter
   setPreference: false # Frontier
-  requirements:
-    - !type:OverallPlaytimeRequirement
-      time: 10800
   startingGear: ReporterGear
   icon: "JobIconReporter"
-  supervisors: job-supervisors-hire # Frontier
+  supervisors: job-supervisors-hop
```
rebaserHEAD: Could it be dropped entirely out. Mostly a nitpick.

Content.Shared/_DV/TapeRecorder/Systems/SharedTapeRecorderSystem.cs
```diff
@@ -0,0 +1,438 @@
...
+    [Dependency] private readonly IRobustRandom _random = default!;
+    [Dependency] private readonly SharedAppearanceSystem _appearance = default!;
+    [Dependency] protected readonly SharedAudioSystem Audio = default!;
+    [Dependency] private readonly SharedDoAfterSystem _doAfter = default!;
+    [Dependency] private readonly ItemSlotsSystem _slots = default!;
+    [Dependency] private readonly SharedUserInterfaceSystem _ui = default!;
+
+    protected const string SlotName = "cassette_tape";
+
+    public override void Initialize()
```
rebaserHEAD, line 36: On the save/load issues in the description: the missing tape visual is because appearance data isn't map-serialized and nothing re-derives it after load. A ComponentStartup subscription here calling UpdateAppearance should fix it. ComponentStartup re-runs when a saved entity loads, while MapInit does not re-fire for entities saved post-init, so don't hang it on MapInit.

Same handler is also the answer to the mode question in the description: reset Mode to Stopped and remove ActiveTapeRecorderComponent there. Both persist right now, so a recorder saved mid-record resumes on load and silently overwrites the tape segment it rolls over.

One caveat: whether the cassette slot contents are already restored by the time startup runs hasn't been verified, so the fix wants an in-game save/load check.

Content.Shared/_DV/TapeRecorder/Systems/SharedTapeRecorderSystem.cs
```diff
@@ -0,0 +1,438 @@
...
+        else
+        {
+            // can't play without a tape in it...
+            if (!TryGetTapeCassette(ent, out _))
+                return;
+
+            EnsureComp<ActiveTapeRecorderComponent>(ent);
+        }
+
+        var sound = ent.Comp.Mode switch
```
rebaserHEAD, line 362: Not all that important, and it's verbatim from Frontier, but FYI: this switch reads the old mode (Mode isn't reassigned until below), so pressing play from stopped plays stop.ogg. Fine to leave for upstream parity, just flagging in case the sounds ever feel backwards in testing.

Content.Server/_DV/TapeRecorder/TapeRecorderSystem.cs
```diff
@@ -0,0 +1,132 @@
...
+        SubscribeLocalEvent<TapeRecorderComponent, PrintTapeRecorderMessage>(OnPrintMessage);
+    }
+
+    /// <summary>
+    /// Given a time range, play all messages on a tape within said range, [start, end).
+    /// Split into this system as shared does not have ChatSystem access
+    /// </summary>
+    protected override void ReplayMessagesInSegment(Entity<TapeRecorderComponent> ent, TapeCassetteComponent tape, float segmentStart, float segmentEnd)
+    {
+        var voice = EnsureComp<VoiceOverrideComponent>(ent);
```
rebaserHEAD, line 39: Minor but: this VoiceOverrideComponent is never removed, so the last replayed speaker's name sticks to the recorder permanently after playback stops (and gets saved with it). Inherited from upstream, fine to leave for parity.

Content.Shared/_DV/TapeRecorder/Systems/SharedTapeRecorderSystem.cs
mnva0: I think I fixed this, but the last line here is kind of unclear

mnva0:
Okay, that should work. I opted to not include a line to remove `ActiveTapeRecorderComponent` because `SetMode` already does that

I feel as if an animation broke but I'm not even that sure if there is one? Something just kind of feels off. I just recall seeing one at some point and now none of the buttons do it. I looked at an old video I made from DeltaV showing off my signal control for tape recorders but those had no animation either

TheRealMasterChief117 (approved):

PR #575, opened 2026-08-29
No Carriers

Eillyrea:
## About the PR

Shipyard consoles no longer save on ships

## Why / Balance

Spawning ships on other ships allows players to load saved ships wherever they want without needing to go to a shipyard console.  
Players needing to travel to get a ship out increases roleplay opportunities visiting stations and interacting with other players instead of sitting in deep space on their ships magically spawning and despawning their saved vessels

Set them to not save so if people really want to get the two small mining vessels from a windreign, they still can, but really, who does that?

## Media

image
image

## Requirements

- [X] I have read the guidelines and documentation relevant to this PR.

## Breaking changes

**Changelog**

:cl:
- remove: Shipyard consoles saving on Ships


TheRealMasterChief117 (changes requested):
Make them inherit from the base class 1 ship contraband since things that are pre mapped to ships shouldn't be illegal

Eillyrea:
The dark deed you have requested is done my lord

rebaserHEAD:
This is fair, some discussion was had but I approve.

TschudyNF:
Nothing is stopping a player from docking at stations and seeking out those roleplay opportunities in the current state.  Additionally, players already park their ships in the middle of nowhere and avoid stations and other players without a carrier console involved.

Removing the ability to operate as a carrier only makes it more difficult for larger ships to operate as they'll need to find parking for their entire size, likely blocking access to other docks, instead being able to deploy a smaller shuttle to make the beer run.

TehFlaminTaco:
This is a bad idea, it just straight up punishes carrier ship designs for the only reason of "forcing people to use the station shipyards", though the only way this does so is... what, once at the start of the shift? When they had to do so anyway to summon the carrier? We don't make roleplay by limiting options.

strace-enjoyer:
i think if people are so hellbent on "evading rp" by using carrier consoles, they'll run to venmar or something, load, and then immediately go back, so this would add rp opportunities which would never be capitalized on, while reducing convenience for legitimate carrier users

Resources/Prototypes/_Mono/Entities/Structures/Machines/Computers/computers_shipyard_mothership.yml
TheRealMasterChief117:
```suggestion
  parent: [BaseStructureDisableToolUse, BaseStructureDestructible, ComputerShipyard, BaseClass1ShipContraband] # Triad - save contra
```

Resources/Prototypes/_Mono/Entities/Structures/Machines/Computers/computers_shipyard_mothership.yml
```diff
@@ -12,13 +12,18 @@
       enum.ShipyardConsoleUiKey.Custom:
         type: ShipyardConsoleBoundUserInterface
   - type: ShipyardListing
+  # Triad Start
+  - type: SavingContraband
+    examineText: ship-saving-contraband-1-text
+  # Triad End
```
TheRealMasterChief117:
```suggestion
```

Resources/Prototypes/_Mono/Entities/Structures/Machines/Computers/computers_shipyard_mothership.yml
```diff
@@ -12,13 +12,18 @@
...
 
 # The empty console
 - type: entity
   id: SellOnlyCarrierComputer
   parent: BaseCarrierComputer
   name: sell only shipyard console
 
+
```
TheRealMasterChief117:
```suggestion
```

TheRealMasterChief117 (approved):

nathan736:
There are currently 6 locations you can already spawn shops from, this pr should have been to just remove the consoles in general, not that i agree with that but its been merged 🤷

PR #594, opened 2026-09-01
SafeCan™! Keeping Your Spacers Safe Since 2828!

rebaserHEAD:
## About the PR
Gas tanks and canisters now carry internal fire suppression: contents that approach plasma ignition temperature while holding flammables get converted to the same mole count of room-temperature water vapour, the valves seize, and any docked tank is fused into the canister. Vessels that still fail (50 atm fragmentation, or destruction while holding gas) turn into a single metal foam block that consumes their contents; empty canisters just break into wreckage. The tank explosion path and canister content dumping are removed entirely, and no gas reaction was touched.

Suppression is keyed on the components, so it covers every GasTank carrier (tanks, jetpacks, gas magboots) and all canister variants, including the ones that redeclare their own Destructible thresholds. Examining a covered vessel shows a bright yellow SafeCan™ label.

## Why / Balance
What if your cans and tanks were actually safe to use! Fire shouldn't be inside of a gas can or gas tank! That goes inside a pipe, silly! Why deal with the old, terribly designed cans and tanks when our proprietary mixture of used coffee grounds and giant space bat saliva can become a frothy mixture that stops ANY pressure from exploding unexpectedly! WOWEE!

## Media

Imagine a slightly moist block of metal foam. That's what happens basically. Now imagine how it smells.

## Requirements
- [x] I have read [Attributing Your Changes](https://github.com/Triad-Sector/Triad_Sector/wiki/Attributing-Your-Changes) and the documentation relevant to this PR.

## How to test
Spawn a filled air tank and VV its volume to 0.01: it should become a foam block, not an explosion. Heat a plasma canister past 100C: contents become water vapour with a popup and the valve seizes. Shoot a charged canister: foam block, nothing vents. Break an empty canister: wreckage only. Unit tests cover the gate and conversion math (6 passing in Content.Tests).

## Breaking changes
The tank fragmentation branch in GasTankSystem.CheckStatus no longer explodes (upstream code commented out, Triad-marked). DumpCanisterBehavior is no longer referenced by any prototype.

**Changelog**
:cl:
- add: Gas tanks and canisters are now equipped with SafeCan™ internal fire suppression, neutralizing flammable contents that approach ignition temperature. Now with MORE used coffee grounds and Giant Space Bat saliva!
- tweak: Overpressurized or destroyed gas vessels turn into a block of metal foam instead of exploding or venting their contents. Keep your Spacers safe from shrapnel! It's SHARP!


Content.Server/_Triad/Atmos/EntitySystems/GasVesselSuppressionSystem.cs
```diff
@@ -0,0 +1,201 @@
...
+        SubscribeLocalEvent<SafeCanLabelComponent, ExaminedEvent>(OnExamined);
+    }
+
+    private void OnExamined(EntityUid uid, SafeCanLabelComponent component, ExaminedEvent args)
+    {
+        if (!args.IsInDetailsRange)
+            return;
+
+        args.PushMarkup(Loc.GetString(component.Label));
+    }
```
TheRealMasterChief117: This event should be moved to shared so that the examine text is predicted between server and client, since you're implementing it as a component. The component itself should also be moved to shared and networked with ``NetworkedComponent` attribute`.

You also don't need to use the initalize here to subscribe to local events anymore, can just use ``SubscribeLocalEvent`` attribute!

Resources/Prototypes/Entities/Structures/Storage/Canisters/gas_canisters.yml
```diff
@@ -6,6 +6,7 @@
   description: A canister that can contain any type of gas. It can be attached to connector ports using a wrench.
   components:
     - type: Animateable
+    - type: SafeCanLabel # Triad: SafeCan examine label
```
TheRealMasterChief117:
```suggestion
```

PR #596, opened 2026-09-02
Radiator Wall Conductance Follows Mass Flux Instead of Saturating

rebaserHEAD:
## About the PR
The radiator's gas-to-body conductance now follows a mass-flux law instead of a clamp: it scales with moles per second through the device as (flow / 100 mol/s)^0.8, capped at 3x, with the same 10% static floor at zero flow. Two new `HeatExchanger` datafields carry the exponent and the cap. Code and guidebook only, no prototype changes.

## Why / Balance
The old flow factor saturated at 10 mol/s, so every pumped loop got the same wall conductance however it was built; the new law lets a dense, fast loop buy more, the way a real exchanger does. In practice this is a fidelity change rather than a balance one: gas already leaves a run of radiators at fin temperature under either law, so loop temperatures barely move and nothing gets nerfed. The new datafields are the knobs for later tuning if the wall is ever made to matter.

## Media
N/A, server-side physics change.

## Requirements

## How to test
Run a hot pumped loop through a radiator chain and confirm it behaves as before; then compare a short run at low and high loop pressure on the same pump setting, where the dense loop should exchange slightly faster. `RadiatorTest` (7, including the new `DenserLoopTransfersMoreHeatThroughTheWall`) and both `PrototypeSaveTest` classes pass locally under the CI build flags.

## Breaking changes
None. Two new `HeatExchanger` datafields (`flowExponent`, `maxFlowFactor`) and `ratedFlow`'s default moves from 10 to 100 mol/s; no prototype in tree declares `ratedFlow`, so nothing pins the old value.

**Changelog**
:cl:
- tweak: Radiator heat exchange now scales with how much gas actually flows through the device instead of saturating at a trickle. Pumped loops behave as before.

### DIFF
--- a/Content.Server/Atmos/Components/HeatExchangerComponent.cs
+++ b/Content.Server/Atmos/Components/HeatExchangerComponent.cs
@@ -59,8 +59,10 @@ public sealed partial class HeatExchangerComponent : Component
 
     /// <summary>
     /// Triad: gas-to-body conductance at rated flow (J/K/sec). Forced
-    /// convection: actual conductance scales between
-    /// <see cref="StaticConductanceFloor"/> and 1x with flow through the device.
+    /// convection: actual conductance scales from
+    /// <see cref="StaticConductanceFloor"/> at zero flow, through 1x at
+    /// <see cref="RatedFlow"/>, up to <see cref="MaxFlowFactor"/> as a
+    /// mass-flux power law (<see cref="FlowExponent"/>).
     /// </summary>
     [ViewVariables(VVAccess.ReadWrite)]
     [DataField("pipeConductance")]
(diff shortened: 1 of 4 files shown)


Content.Server/Atmos/Components/HeatExchangerComponent.cs
```diff
@@ -76,12 +78,33 @@ public sealed partial class HeatExchangerComponent : Component
...
-    public float RatedFlow = 10f;
+    public float RatedFlow = 100f;
+
+    /// <summary>
+    /// Triad: exponent of the mass-flux law. Real gas-side film coefficients
+    /// go as mass flux^0.8 in turbulent flow (Dittus-Boelter), so doubling
+    /// the flow buys ~1.74x conductance, not 2x.
+    /// </summary>
+    [ViewVariables(VVAccess.ReadWrite)]
+    [DataField("flowExponent")]
```
TheRealMasterChief117: View variables read write is already implemented on datafield attribute

Datafield names that are equal to the datafield itself are also unneeded

